PR #1, opened 2023-02-21
Upgrade airflow; load environment in dag files

moradology:
This PR fixes issues around deployment encountered after upgrading MWAA to the newest version, applies some new IAM policies around ECS usage within airflow, implements a `register-use-deregister` strategy for ECS tasks (as is used in CSDA), and updates the environment loading logic so that it works nicely with `handler` code that is loaded via submodule


moradology:
Closed as subsumed by other open PRs

PR #2, opened 2023-02-22
Fix collections that have intermittent data

sbquinlan:
Fixes NASA-IMPACT/veda-data-pipelines#221 and 
Fixes NASA-IMPACT/veda-data-pipelines#281 at the configuration level in the repo.


j08lue (approved):
Approving changes here but it would be great to close the periodicity issue only when this change was propagated to the STAC and validated in the Dashboard.

PR #5, opened 2023-02-27
Creating a task definition for VEDA stac build

amarouane-ABDELHAK:

Previously we were doing register the task then run `build stac` then de-register the task but it turns out it added so much overhead. This implementation will use one AWS task definition to run `build stac` 

## Changes

* Removing register task from processing dag
* Adding an AWS task definition for `build stac` task
* Adding random uuid to run DAG (in case we want to re-run a successful task) 

## PR Checklist

- [x] PR comment added

### DIFF
--- a/dags/example_dag.py
+++ b/dags/example_dag.py
@@ -1,7 +1,7 @@
-import pendulum
 import logging
 import time
 
+import pendulum
 from airflow import DAG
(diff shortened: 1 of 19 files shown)


slesaad (approved):
lgtm

PR #15, opened 2023-03-09
Vector ingest DAG and branching

smohiudd:
This PR includes a new vector/raster branch in discover as well as vector ingest DAG.

image

One issue that needs to be addressed: the vector DB is in another VPC so will need to sort out a security group and ingress rule.


docker_tasks/vector_ingest/Dockerfile
```diff
@@ -1,19 +1,9 @@
-FROM --platform=linux/amd64 ghcr.io/lambgeo/lambda-gdal:3.5-python3.9
+FROM --platform=linux/amd64 python:3.9
```
smohiudd, line 1: This was using a gdal specific lambda image. Will eventually have to replace with another gdal image.

amarouane-ABDELHAK: What `cmd` are you using that requires `gdal`?

amarouane-ABDELHAK: I see you are running `ogr2ogr`

docker_tasks/vector_ingest/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM --platform=linux/amd64 osgeo/gdal:alpine-normal-latest
```
smohiudd, line 1: @ranchodeluxe we're using Gdal 3.7 so we're able to use the "upsert" instead of "overwrite" if we choose to.

dags/veda_data_pipeline/veda_process_vector_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+    "upload": "<false> | true",
+    "cogify": "false | true",
+    "payload": "<s3_uri_event_payload",
+}
+dag_args = {
+    "start_date": pendulum.today("UTC").add(days=-1),
+    "schedule_interval": None,
+    "catchup": False,
+    "doc_md": dag_doc_md,
+}
```
moradology, line 50: Perhaps worth updating this, as by this point we're surely dealing with only vectors (json/gdb/etc, I assume)

infrastructure/variables.tf
```diff
@@ -48,3 +48,12 @@ variable "min_workers" {
     production = 3
   }
 }
+variable "vector_secret_name" {
```
moradology, line 51: Perhaps this should be `postgres_secret_name` to correspond to how we reference it elsewhere

moradology (approved):
One minor suggestion. Otherwise, looks good to me. Perhaps @amarouane-ABDELHAK could double check the TF work here (it looks right to me but my certainty level there is a bit lower)

infrastructure/variables.tf
smohiudd: I agree. The variable names are a little confusing. Will do a once over and change them .

infrastructure/task_definition.tf
```diff
@@ -16,12 +16,43 @@ resource "aws_ecs_task_definition" "veda_task_definition" {
...
+  container_definitions = jsonencode([
+
+    {
+      name      = "${var.prefix}-veda-vector_ingest"
+      image     = "${local.account_id}.dkr.ecr.${local.aws_region}.amazonaws.com/${var.prefix}-veda-vector_ingest"
+      essential = true,
+      logConfiguration = {
+        "logDriver": "awslogs",
+        "options": {
+          "awslogs-group": module.mwaa.log_group_name,
```
ranchodeluxe, line 43: non-blocking question for @amarouane-ABDELHAK : noticed that `module.mwaa.source` is pointing to a private repo. Can we move that into "NASA-IMPACT" by chance?

amarouane-ABDELHAK: Yes, I am planning to open a PR to point it to NASA-IMPACT after this PR

amarouane-ABDELHAK:
I got this error when I tried to test this deployment
`│ Error: Invalid index
│
│   on main.tf line 94, in resource "local_file" "mwaa_variables":
│   94:       vector_subnet_1 = data.aws_subnets.private.ids[0]
│     ├────────────────
│     │ data.aws_subnets.private.ids is empty list of string
│
│ The given key does not identify an element in this collection value: the collection has no elements.
╵
╷
│ Error: Invalid index
│
│   on main.tf line 95, in resource "local_file" "mwaa_variables":
│   95:       vector_subnet_2 = data.aws_subnets.private.ids[1]
│     ├────────────────
│     │ data.aws_subnets.private.ids is empty list of string
│
│ The given key does not identify an element in this collection value.
╵
Releasing state lock. This may take a few moments...`

amarouane-ABDELHAK (approved):
Some net picking maybe we should move the `data` to fetch the subnets ids to `aws_data.tf` file to have all the data related resources in the same file

PR #28, opened 2023-04-07
add back the correct refresh domain

ranchodeluxe:

* add back the correct refresh domain that somehow got removed from `main`
* add pre-commit hook yaml for black

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Run linters
         uses: wearerequired/lint-action@v2
         with:
-          continue_on_error: true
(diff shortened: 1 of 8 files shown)


ividito (approved):

anayeaye (approved):
Second the approval, that path definitely exists. Would love to see this in configuration instead at some point but that's a style preference.

ranchodeluxe:
> Second the approval, that path definitely exists. Would love to see this in configuration instead at some point but that's a style preference.

yeah, good point. it should be a config value. we can do that later

PR #35, opened 2023-04-25
Feature/lf archival changes

ranchodeluxe:

Addresses [veda-analytics#48](https://github.com/orgs/NASA-IMPACT/projects/51/views/6?pane=issue&itemId=26015087) and ingest large fire archival inputs into collections

## Changes

include `ogr2ogr` commands for large fire inputs

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/docker_tasks/vector_ingest/handler.py
+++ b/docker_tasks/vector_ingest/handler.py
@@ -124,7 +124,6 @@ def load_to_featuresdb(filename: str, collection: str):
                 check=True,
                 capture_output=True,
             )
-            alter_datetime_add_indexes(collection)
         elif collection == "perimeter":
             subprocess.run(
                 [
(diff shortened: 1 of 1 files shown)


docker_tasks/vector_ingest/handler.py
```diff
@@ -145,11 +144,35 @@ def load_to_featuresdb(filename: str, collection: str):
...
+                    "-progress",
+                ],
+                check=True,
+                capture_output=True,
+            )
         else:
             print("Not a valid fireline collection")
             return {"status": "failure"}
 
+        alter_datetime_add_indexes(collection)
```
smohiudd, line 174: If you you get to else condition (i.e. "Not a valid fireline collection"), then the `alter_datetime_add_indexes` will fail since it's not passing a collection

smohiudd (commented):
Added a comment about `alter_datetime_add_indexes(collection)`.

docker_tasks/vector_ingest/handler.py
ranchodeluxe: line 172 has the `return` there that short circuits this right?

smohiudd: Yes you are correct!

smohiudd (approved):

PR #49, opened 2023-04-28
use count to only provision RDS in dev env

amarouane-ABDELHAK:

Only provision RDS cluster in dev


smohiudd (approved):

ranchodeluxe:
let's rock and roll 🥳

PR #51, opened 2023-04-28
Allow Developers to `revert` from the UI in Priority-0 Situations

ranchodeluxe:
### Summary

Fixes first AC of: https://github.com/NASA-IMPACT/veda-analytics/issues/64

Let's talk in tag-up through the implications of tightly coupling IAC and CI/CD b/c now a `revert` semantically means more about code+infra and we can end up in a bad state. I think there's a better practice at hand

### Changes

* allow revert branches to target `main` and `dev`

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -10,10 +10,17 @@ jobs:
       runs-on: ubuntu-latest
       steps:
         - name: Check branch
-          if: github.base_ref == 'main' && github.head_ref != 'dev' || github.base_ref == 'production' && github.head_ref != 'main'
           run: |
-            echo "ERROR: You can only merge to main from dev and to production from main"
-            exit 1
+            if [[ $GITHUB_BASE_REF == "main" && ($GITHUB_HEAD_REF != "dev" || $GITHUB_HEAD_REF != "revert-"*) ]];then
+              echo "ERROR: You can only merge to 'main' from 'dev' or a 'revert-*' branch"
+              exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd.yml
amarouane-ABDELHAK, line 1: From ChatGPT:
Screen Shot 2023-05-01 at 9 21 05 AM
But for a better refactored code I would suggest
`- name: Check branch
  run: |
    if [[ $GITHUB_BASE_REF == "main" ]]; then
      if [[ $GITHUB_HEAD_REF != "dev" && $GITHUB_HEAD_REF != "revert-"* ]]; then
        echo "ERROR: You can only merge to 'main' from 'dev' or a 'revert-*' branch"
        exit 1
      fi
    elif [[ $GITHUB_BASE_REF == "production" ]]; then
      if [[ $GITHUB_HEAD_REF != "main" && $GITHUB_HEAD_REF != "revert-"* ]]; then
        echo "ERROR: You can only merge to 'production' from 'main' or a 'revert-*' branch"
        exit 1
      fi
    fi`
Since we only run the check if the base_ref branch is either "main" or "production"
thoughts?

ranchodeluxe: nice catch ChatGPT! yeah, that refactored big works too, let me change

amarouane-ABDELHAK (approved):

PR #79, opened 2023-06-08
Fix items dir reference

smohiudd:
When running `./scripts/item.sh` the items were not being discovered because the directory reference was incorrect. This PR fixes the directory name.


amarouane-ABDELHAK (approved):

slesaad:
Doesn't make sense anymore

PR #110, opened 2024-02-21
Remove legacy stuff

slesaad:
## Description
Removes legacy stuff that no longer make sense

## Changes
- Removes data folder
- Removes collection .sql files
- Removes old script that referenced the data/ folder (documentation to make the requests exist in wiki and docs)


anayeaye (approved):
Approve with a comment and a question
- With this cleanup everything in OPERATING.md can go except maybe the glossary could get tucked into the main README?
- We don't use the python from scripts/ anymore? All of these scripts are unused duplicate of the code in docker_tasks?

slesaad:
@anayeaye thanks for reminding. I deleted `OPERATING.md` because it's mostly outdated information at this point. We can probably create an updated one down the line.

PR #111, opened 2024-02-27
Update for mono repo

slesaad:
# Description
Changes to work with `veda-deploy` + others discovered along the way

# Changes
- Added composite deploy action that can be used by `veda-deploy` as well as `cicd` in this repo
- Made a lot of env vars optional (assume roles, vector info, etc)
- Update env var names to work together with other repo for mono repo deployment - https://github.com/NASA-IMPACT/veda-data-airflow/issues/112
- Bring in the fix for https://github.com/US-GHG-Center/ghgc-data-airflow/issues/20, from ghgc project https://github.com/US-GHG-Center/ghgc-data-airflow/pull/19
- Added support for permissions boundary for MCP
- Added secret access to workflows api lambda

# Prerequisites
https://github.com/NASA-IMPACT/veda-auth/pull/27

# TODO
- [x] Permission to workflows lambda to read cognito secret

PR #115, opened 2024-03-05
Remove features API, add veda deploy action

smohiudd:
PR includes the following:

- The current implementation of veda features API requires a separate VPC which will not work in MCP. There is a [PR in veda-features-api](https://github.com/NASA-IMPACT/veda-features-api/pull/27/files) which uses a shared VPC but has not been merged. Until veda-features-api uses a shared VPC, this PR comments out the features api in airflow so it can be deployed in MCP. 
- Includes deploy action that can be used by `veda-deploy`


.github/actions/terraform-deploy/action.yml
```diff
@@ -0,0 +1,62 @@
+name: Deploy
```
slesaad, line 1: this is already in https://github.com/NASA-IMPACT/veda-data-airflow/pull/111 so can we not add it here?

smohiudd: yeah we can probably remove this one (it's also out of date).

PR #116, opened 2024-03-05
fix: coalesce empty string to false

ividito:
Fixes deployment failures if CLOUDFRONT_ID is using default value (empty string).


smohiudd:
@ividito to confirm, this is for the conditional for if the cloudfront behaviour should be updated? So right now we're not going to worry about this and just use the apigateway url?

botanical (commented):
The change looks technically valid to me! I'm going to test it with my MCP deployment. But I'll let someone else give the final approval

slesaad (approved):

PR #118, opened 2024-03-11
Fix workflows publish collections

smohiudd:
There are two changes in this PR:

- `requests` POST expects the `data` argument when passing JSON instead of `json` which is for dicts. I tried exported the pydantic model to dict but there was an issue with datetime serialization. 
- The `get_token()` function was returning an instance of [HTTPBearer](https://fastapi.tiangolo.com/reference/security/#fastapi.security.HTTPBearer) but we need pass just the credentials to the publish endpoint


ividito (approved):

ividito:
> I tried exported the pydantic model to dict but there was an issue with datetime serialization

I think (need to double check) that this will be resolved by the ingest API's validations.

slesaad (approved):

PR #154, opened 2024-05-23
Remove filename extension

smohiudd:
Removes only the last extension from the filename and uses this as the item id.

### DIFF
--- a/dags/veda_data_pipeline/utils/s3_discovery.py
+++ b/dags/veda_data_pipeline/utils/s3_discovery.py
@@ -111,12 +111,10 @@ def construct_single_asset_items(discovered_files: List[str]) -> dict:
     for uri in discovered_files:
         # Each file gets its matched asset type and id
         filename = uri.split("/")[-1]
-        filename_without_extensions = Path(filename)
-        while filename_without_extensions.suffix: #for filenames that have multiple extensions i.e. *.tif.cog
-             filename_without_extensions=filename_without_extensions.with_suffix('')
+        filename_without_extension = Path(filename).stem
         prefix = "/".join(uri.split("/")[:-1])
         item = {
-            "item_id": str(filename_without_extensions),
+            "item_id": filename_without_extension,
             "assets": {
                 "default": {
                     "title": "Default COG Layer",


dags/veda_data_pipeline/utils/s3_discovery.py
```diff
@@ -111,12 +111,10 @@ def construct_single_asset_items(discovered_files: List[str]) -> dict:
...
-        filename_without_extensions = Path(filename)
-        while filename_without_extensions.suffix: #for filenames that have multiple extensions i.e. *.tif.cog
-             filename_without_extensions=filename_without_extensions.with_suffix('')
+        filename_without_extension = Path(filename).stem
         prefix = "/".join(uri.split("/")[:-1])
         item = {
-            "item_id": str(filename_without_extensions),
+            "item_id": filename_without_extension,
             "assets": {
                 "default": {
```
anayeaye, line 119: I hadn't noticed this earlier and it is not part of the change proposed in this PR, but: does this mean that when more specific asset info in the discovery config we are making items with an asset key `"default"` instead of `"cog_default"`?  Recent ingests to dev still produce items with cog_default assets so this is probably not a relevant comment

anayeaye (approved):

dags/veda_data_pipeline/utils/s3_discovery.py
```diff
@@ -111,12 +111,10 @@ def construct_single_asset_items(discovered_files: List[str]) -> dict:
     for uri in discovered_files:
         # Each file gets its matched asset type and id
         filename = uri.split("/")[-1]
-        filename_without_extensions = Path(filename)
-        while filename_without_extensions.suffix: #for filenames that have multiple extensions i.e. *.tif.cog
-             filename_without_extensions=filename_without_extensions.with_suffix('')
+        filename_without_extension = Path(filename).stem
```
botanical, line 114: Do we care about the collection `modis-fire-anomalies-diff-covid-19-changing-landscape`? The item feature.id string `MODIS-AF_anomalies_USA_202010_M10_0.25-dashboard` would be `MODIS-AF_anomalies_USA_202010_M10_0`. Is that okay, since we might not need to support that collection? 🤔 

https://staging-stac.delta-backend.com/collections/modis-fire-anomalies-diff-covid-19-changing-landscape/items

dags/veda_data_pipeline/utils/s3_discovery.py
smohiudd: @anayeaye I think this is the where the `default` gets changed to `cog_default`: https://github.com/NASA-IMPACT/veda-data-airflow/blob/62ac7b1ae0fab843694fd1b2ef8d4066d5bf0cac/docker_tasks/build_stac/utils/stac.py#L120

dags/veda_data_pipeline/utils/s3_discovery.py
smohiudd: I had no idea we had so many of weird filenames! Are we not supporting it or it's not used in the dashboard currently?

botanical: from a cursory search it doesn't look like it's used in the dashboard, but I could be wrong! @anayeaye , do you know?

botanical: https://github.com/NASA-IMPACT/veda-config/tree/develop/datasets I'm guessing since it's not here, we're good

botanical (approved):

PR #155, opened 2024-05-23
feat: support for scheduled ingest triggers

slesaad:
**Summary:** Adds support for scheduled ingest

## Changes

* a new `generate_dags.py` file that generates a separate dag for each collection that has a schedule - so that we can run the ingests based on different schedules
* updated the s3 discovery such that if it is run on a schedule, it'll only grab files that have been updated since the last successful run

See more - https://github.com/US-GHG-Center/ghgc-data-airflow/pull/28


dags/requirements.txt
```diff
@@ -1,4 +1,4 @@
---constraint /usr/local/airflow/dags/requirements-constraints.txt
+#--constraint /usr/local/airflow/dags/requirements-constraints.txt
```
ividito, line 1: I'm not sure why this change is needed - found the source at https://github.com/US-GHG-Center/ghgc-data-airflow/pull/27 but I just want to double check that we're still pulling in the constraints elsewhere, or that we're certain that they're no longer needed.

amarouane-ABDELHAK: I recall removing the constraint because it caused an issue with a specific version (possibly rasterio) which makes the scheduler to fail, but this change seems to not have any side effect.

ividito (approved):

PR #156, opened 2024-05-24
fix: make assume role optional

slesaad:

Assume role is supposed to be optional.

## Changes

Make assume role optional


smohiudd:
@slesaad I believe the `BACKEND_VPC_ID` was added because in UAH we have two VPCs - one for airflow and one for everything else. I agree we should probably have one VPC and will need to redeploy dev to use the `veda-shared-base/network/vpc` VPC and exclude the airflow VPC

infrastructure/main.tf
```diff
@@ -224,10 +181,7 @@ resource "aws_security_group" "workflows_api_handler_sg" {
   lifecycle { create_before_destroy = true }
 }
 
-# Function to build the JWKS URL
-locals {
-  build_jwks_url = "${format("https://cognito-idp.%s.amazonaws.com/%s/.well-known/jwks.json", local.aws_region, var.userpool_id)}"
-}
```
smohiudd, line 230: Is this not used anywhere? that's why it was removed?

infrastructure/locals.tf
```diff
@@ -0,0 +1,53 @@
...
+
+    conditional_workflows_lambda_policy = var.data_access_role_arn != "" ? concat(local.workflows_lambda_policy, [
+        {
+        Effect: "Allow",
+        Action: ["sts:AssumeRole"],
+        Resource: [var.data_access_role_arn]
+        }
+    ]) : local.workflows_lambda_policy
+
+  build_jwks_url = "${format("https://cognito-idp.%s.amazonaws.com/%s/.well-known/jwks.json", local.aws_region, var.userpool_id)}"
```
slesaad, line 52: `build_jwks_url` was moved here @smohiudd ☝️

infrastructure/main.tf
slesaad: https://github.com/NASA-IMPACT/veda-data-airflow/pull/156#discussion_r1613733005

slesaad:
> @slesaad I believe the `BACKEND_VPC_ID` was added because in UAH we have two VPCs - one for airflow and one for everything else. I agree we should probably have one VPC and will need to redeploy dev to use the `veda-shared-base/network/vpc` VPC and exclude the airflow VPC

hmm - then let me update this to use `VPC_ID` as the default for `BACKEND_VPC_ID`

infrastructure/locals.tf
smohiudd: oh missed that. Thanks!

smohiudd (approved):

ividito (approved):

PR #159, opened 2024-05-31
feat: extract dataset publish

jjfrench:

Addresses [VEDA-XX: Develop amazing new feature](https://github.com/NASA-IMPACT/veda-data-pipelines/issues/XX)

## Changes

* DAG based off the /dataset/publish endpoint in the workflow api
  * Checks that a collection doesn't already exists, generates a new template collection
  * Override basic extents with spatial_extent and temporal_extent in dataset config
  * Kicks off discovery DAG after ensuring that a collection exists

## PR Checklist
- [ ] Integration tests


ividito (approved):
Co-reviewed with @jjfrench, since I contributed some bugfixes

dags/veda_data_pipeline/utils/collection_generation.py
```diff
@@ -0,0 +1,135 @@
...
+            **{
+                key: dataset[key]
+                for key in GenerateCollection.common_fields
+                if key in dataset.keys()
+            },
+        }
+
+        # Default REQUIRED fields
+        if not collection_dict.get("title"):
+            collection_dict["description"] = dataset["collection"]
```
anayeaye, line 41: should this be `collection_dict["title"] = dataset["collection"]`?

dags/veda_data_pipeline/utils/collection_generation.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class GenerateCollection:
+    common_fields = [
+        "title",
+        "description",
+        "license",
+        "links",
+        "time_density",
+        "is_periodic",
```
anayeaye, line 16: by using a custom model, will other fields be stripped from the input if not included here? Or will these be dropped by the custom schema anyway? (please disregard if this is something we have to live with until we transition to pystac).

`[ renders, stac_extensions, assets, item_assets, summaries, and providers]` are all valid fields that could be provided for a new dataset

dags/veda_data_pipeline/utils/collection_generation.py
```diff
@@ -0,0 +1,135 @@
...
+                            [
+                                x.isoformat().replace("+00:00", "Z")
+                                for x in list(temporal_extent)
+                            ]
+                        ]
+                    },
+                }
+            )
+
+        collection_stac["item_assets"] = {
```
anayeaye, line 111: this isn't a required stac field so can only add it when provided in the request?

dags/veda_data_pipeline/utils/collection_generation.py
jjfrench: `title` isn't a required field but `description` is

anayeaye: Ah, that makes sense then. Should it be `if not collection_dict.get("description")` instead of title?

dags/veda_data_pipeline/utils/collection_generation.py
ividito: I think we still want to include this as a default for `/dataset` inputs, since the generated defaults will result in a minimal dashboard-ready collection. For collections that deliberately have no `item_assets`, or have multiple assets, we should probably point users to the `/ingest/collections` endpoint.

dags/veda_data_pipeline/utils/collection_generation.py
jjfrench: mmhmm mmhmm, :nod:

dags/veda_data_pipeline/utils/collection_generation.py
jjfrench: Isayah and I talked this over, we added these to the common_fields but don't necessarily think we need to add in validation for these fields in this PR. Such as the renders stac_extension should be listed if the fields are present, would be something handled in a different PR or service

I think there could potentially be a larger list if we start throwing in all the stac_extensions?

anayeaye: > don't necessarily think we need to add in validation for these fields in this PR

100% agree, I am just hoping to prevent stripping valid information that an operator might have put in the collection record (people are surprised when they take the time to write renders metadata only to see it has disappeared in the stac api after running the old datasets endpoint). I will be curious to see if adding in these fields will allow users to submit more detailed collection metadata.

> I think there could potentially be a larger list if we start throwing in all the stac_extensions?

Also agree, I think eventually we need to be more flexible and use something like the pystac Collection instead of hard coded pydantic models.

dags/veda_data_pipeline/utils/collection_generation.py
anayeaye: Personally, I don't think it is helpful to have a collection item_assets definition of cog_default in all collections by default but I don't think it poses much risk at this point when we do not have algorithmic parsing of collection metadata that would fail when collections falsely advertise cog_default assets. We do risk user confusion/mistrust of collection metadata so maybe we can revisit this on the next round.

dags/veda_data_pipeline/utils/collection_generation.py
anayeaye: I guess it will now be possible for someone to submit invalid metadata in a collection as in `providers: None`.

The datasets endpoint is a can of 🪱 s; sorry I nitpicked it.

smohiudd: I like the idea of pystac collection as a validation method for collections. Could be something to explore. @jjfrench if you want to set up a meeting we can all discuss!

ividito: After our discussion earlier, I changed this behaviour to exclude invalid fields, rather than to only include specified common fields. Also note parallel work being done on GHG in #179 

I think we can break out `pystac` integration to a separate issue, either to be added as a step in Airflow or as a validator in `veda-backend`.

dags/veda_data_pipeline/utils/collection_generation.py
```diff
@@ -0,0 +1,131 @@
...
+        "stac_version": "1.0.0",
+    }
+    keys_to_ignore = ["collection", "data_type", "sample_files", "discovery_items", "spatial_extent", "temporal_extent", "is_periodic", "time_density"]
+
+            
+    def get_template(self, dataset: Dict[str, Any]) -> dict:
+        extra_fields = {
+                key: dataset[key]
+                for key in dataset.keys()
+                if (key not in GenerateCollection.common_fields) and (key not in GenerateCollection.keys_to_ignore)
```
anayeaye, line 26: Is it common not common_fields here?
```
 if (key not in GenerateCollection.common) and (key not in GenerateCollection.keys_to_ignore)
```

anayeaye (approved):
Approve with known 'datasets' model issues:
- airflow dags use properties that are also part of the stac spec so collections with assets (like thumbnails) cannot be ingested via the dataset model
- datetime configs not tested manually
- multi asset items not tested manually
- the datasets model in airflow DAG is diverging from the datasets model in the workflows-api but this is known and the plan is to eventually connect the workflows-api datasets endpoints to the datasets DAG

PR #167, opened 2024-06-17
fix: add default root path in config directly

botanical:

Addresses [VEDA-164: Fix origin and distribution api/workflow behavior to cloudfront distribution](https://github.com/NASA-IMPACT/veda-data-pipelines/issues/164)

## Changes

* add default root path 
* make origin naming consistent with backend pattern
* update the viewer policy to work with cloudfront

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


infrastructure/terraform.tfvars.tmpl
```diff
@@ -16,7 +16,7 @@ stac_url="${VEDA_STAC_URL}"
 vector_secret_name="${VECTOR_SECRET_NAME}"
 vector_security_group="${VECTOR_SECURITY_GROUP}"
 vector_vpc="${VECTOR_VPC:-null}"
-workflow_root_path="${WORKFLOW_ROOT_PATH}"
+workflow_root_path="${WORKFLOW_ROOT_PATH:-/api/workflows}"
```
amarouane-ABDELHAK, line 19: This line ensures that the Terraform variable workflow_root_path will default to "/api/workflows" if WORKFLOW_ROOT_PATH is not defined in AWS Secret Manager. The only drawback is that the default value for workflow_root_path defined in variables.tf will always be ignored.
Probably a better solution is to change the name of the variable defined in AWS Secret Manager to TF_VAR_workflow_root_path and remove the line workflow_root_path="${WORKFLOW_ROOT_PATH:-/api/workflows}" from infrastructure/terraform.tfvars.tmpl. This way, Terraform will use the default value in variables.tf if TF_VAR_workflow_root_path is not defined in AWS Secret Manager.

botanical:
Tested using mcp-test and verified that the changes work

workflows_api/runtime/src/main.py
```diff
@@ -30,7 +30,7 @@
         "url": "https://www.apache.org/licenses/LICENSE-2.0.html",
     },
     contact={"url": "https://github.com/NASA-IMPACT/veda-backend"},
-    root_path=settings.workflow_root_path,
+    root_path=settings.workflow_root_path or "/api/workflows",
```
smohiudd, line 33: A little nervous about this since we usually keep defaults in config. But seems that there isn't a simple fix to handle empty stings

botanical:
Going to handle overall veda routes next PI

PR #172, opened 2024-06-20
(fix): non scheduled discovery and dag import error

slesaad:
# Changes
- Fix for "Dag import error" when there's no scheduled collection ingests
- Fix for s3 files not found when it's not a scheduled discovery

### DIFF
--- a/dags/generate_dags.py
+++ b/dags/generate_dags.py
@@ -2,6 +2,7 @@
 Builds a DAG for each collection (indicated by a .json file) in the <BUCKET>/collections/ folder.
 These DAGs are used to discover and ingest items for each collection.
 """
+
 from airflow.models.variable import Variable
 
 
@@ -18,7 +19,7 @@ def generate_dags():
     client = boto3.client("s3")
     response = client.list_objects_v2(Bucket=bucket, Prefix="collections/")
 
-    for file_ in response["Contents"]:
+    for file_ in response.get("Contents", []):
         key = file_["Key"]
         if key.endswith("/"):
             continue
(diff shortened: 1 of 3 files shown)


dags/generate_dags.py
```diff
@@ -18,7 +19,7 @@ def generate_dags():
     client = boto3.client("s3")
     response = client.list_objects_v2(Bucket=bucket, Prefix="collections/")
 
-    for file_ in response["Contents"]:
+    for file_ in response.get("Contents", []):
```
slesaad, line 22: If there are no files in the directory, there's no "Contents" in response, which causes the "DAG import Errors (1)"
image

This fixes that.

dags/veda_data_pipeline/groups/discover_group.py
```diff
@@ -22,7 +22,7 @@ def discover_from_s3_task(ti, event={}, **kwargs):
         **ti.dag_run.conf,
     }
     last_successful_execution = kwargs.get("prev_start_date_success")
-    if last_successful_execution:
+    if event.get("schedule") and last_successful_execution:
         config["last_successful_execution"] = last_successful_execution.isoformat()
```
slesaad, line 26: We only want to add the `last_successful_execution` field to the event if it is a scheduled delivery (which is determined by if there's a "schedule" field in the incoming `event`.

This right now is causing "no files discovered" error, because the files are there since the `veda_discover` dag was last run.

This fixes this.

docker_tasks/build_stac/requirements.txt
```diff
@@ -3,7 +3,7 @@ awslambdaric
 boto3
 pystac==1.4.0
 python-cmr
-rasterio==1.3.6
+rasterio==1.3.10
```
slesaad, line 6: to fix the following error:

```
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject
```

botanical (approved):

sanzog03 (approved):

PR #185, opened 2024-07-10
json prepared for ingest api now allows arbitrary key value pairs

sanzog03:

Addresses [sub-feature of GHGC-269: Fix workflows api for GHGC](https://github.com/US-GHG-Center/ghgc-architecture/issues/269)

## Changes

* Allow arbitrary key value pairs in the request body payload

### DIFF
--- a/workflows_api/runtime/src/collection_publisher.py
+++ b/workflows_api/runtime/src/collection_publisher.py
@@ -170,6 +170,11 @@ class Publisher:
                 }
             }
 
+        # if the dataset has the attributes provided already in the request json, use them instead of the generated ones
+        for key, _ in collection_stac.items():
+            if key in dataset_dict:
+                collection_stac[key] = dataset_dict[key]
+
         return collection_stac
 
     def generate_stac(
--- a/workflows_api/runtime/src/collection_publisher.py
+++ b/workflows_api/runtime/src/collection_publisher.py
@@ -58,21 +58,25 @@ class Publisher:
 
     def __init__(self) -> None:
         self.func_map = {
-            DataType.zarr: self.create_zarr_collection,
-            DataType.cog: self.create_cog_collection,
+            DataType.zarr.value: self.create_zarr_collection,
+            DataType.cog.value: self.create_cog_collection,
         }
 
     def get_template(self, dataset: Union[ZarrDataset, COGDataset]) -> dict:
         dataset_dict = dataset.dict()
-        collection_dict = {
-            "id": dataset_dict["collection"],
-            **Publisher.common,
-            **{
-                key: dataset_dict[key]
-                for key in Publisher.common_fields
-                if key in dataset_dict.keys()
-            },
-        }
+        collection_dict = dataset_dict
+        # manage data as needed by ingest
+        collection_dict["id"] = dataset_dict["collection"]
+        collection_dict.update(Publisher.common)
+        collection_dict.update({
+            key: dataset_dict[key]
+            for key in Publisher.common_fields
+            if key in dataset_dict.keys()
+        })
+        # Convert enum values to their string equivalents
+        for key, value in dataset_dict.items():
+            if isinstance(value, DataType):
+                dataset_dict[key] = value.value
         return collection_dict
 
     def _create_zarr_template(self, dataset: ZarrDataset, store_path: str) -> dict:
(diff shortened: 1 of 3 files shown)


workflows_api/runtime/src/schemas.py
```diff
@@ -181,6 +182,9 @@ def object_is_accessible(cls, values):
         )
         return values
 
+    class Config:
+        extra = Extra.allow
+
```
slesaad, line 187: this shouldn't need extras

workflows_api/runtime/src/schemas.py
```diff
@@ -262,6 +268,8 @@ def check_sample_files(cls, values):
             )
         return values
 
+    class Config:
+        extra = Extra.allow
```
slesaad, line 272: does this not get inherited from `Dataset` ?

workflows_api/runtime/src/schemas.py
```diff
@@ -280,3 +288,6 @@ def only_one_discover_item(cls, discovery_items):
                 "Zarr dataset should include zarr_store in its discovery item"
             )
         return discovery_items
+
+    class Config:
+        extra = Extra.allow
```
slesaad, line 293: does this not get inherited from `Dataset` ?

workflows_api/runtime/src/collection_publisher.py
```diff
@@ -58,21 +58,25 @@ class Publisher:
 
     def __init__(self) -> None:
         self.func_map = {
-            DataType.zarr: self.create_zarr_collection,
-            DataType.cog: self.create_cog_collection,
+            DataType.zarr.value: self.create_zarr_collection,
+            DataType.cog.value: self.create_cog_collection,
```
slesaad, line 62: 🤔

workflows_api/runtime/src/collection_publisher.py
```diff
@@ -58,21 +58,25 @@ class Publisher:
...
+        collection_dict.update(Publisher.common)
+        collection_dict.update({
+            key: dataset_dict[key]
+            for key in Publisher.common_fields
+            if key in dataset_dict.keys()
+        })
+        # Convert enum values to their string equivalents
+        for key, value in dataset_dict.items():
+            if isinstance(value, DataType):
+                dataset_dict[key] = value.value
```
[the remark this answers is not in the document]
sanzog03: there was a key value namely `datatype: <Datatype.cog>`
It was not a valid json for the ingest api call

workflows_api/runtime/src/schemas.py
sanzog03: makes sense. removing it.

workflows_api/runtime/src/schemas.py
sanzog03: makes sense. removing it.

workflows_api/runtime/src/schemas.py
sanzog03: This one is for `S3Input` and if we need to allow extra attributes inside discovery_items, we need this as well.

workflows_api/runtime/src/schemas.py
sanzog03: makes sense. Removing it!

workflows_api/runtime/src/schemas.py
sanzog03: makes sense. Removing it!

workflows_api/runtime/src/collection_publisher.py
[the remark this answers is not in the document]
sanzog03: The value for the key `data_type` in the collection is an enumerator. Its value will be `<DataType.cog>` by default, which is not a valid JSON and will cause an issue during API call for ingest. Previously, this was ignored in the JSON body for ingest, However, as we will allow everything now, this will come in as well. So fixed it here.

workflows_api/runtime/src/collection_publisher.py
sanzog03: the expected map is between string key and function value. It was rather <Datatype.cog> as key and function as value.

sanzog03 (commented):
replies to the previous comments

workflows_api/runtime/src/schemas.py
slesaad: we don't need to allow extra attributes inside `discovery_items`

slesaad (approved):

PR #190, opened 2024-07-12
Add Discover short circuit to discover_from_s3 airflow task

amarouane-ABDELHAK:
In case discover_from_s3 didn't discover any file form S3 (list is empty) the DAG will gracefully skip all the remaining tasks instead of failing.


amarouane-ABDELHAK:
Case of no file discovered
image
Case of files discovered
image

anayeaye (approved):

Issue #195, opened 2024-07-22
Necessary values are propagated from veda-data-airflow to veda-routes repo for CloudFront distribution origin management


botanical:
## Context

- Epic Link: https://github.com/NASA-IMPACT/veda-architecture/issues/469
- ADR: https://github.com/NASA-IMPACT/veda-architecture/pull/464

Propagate the necessary values for CloudFront distribution origin management 
- [workflows api ](https://github.com/NASA-IMPACT/veda-data-airflow/blob/dev/infrastructure/outputs.tf#L13-L15)

## Acceptance Criteria

- [x] API Gateway value for workflows API is propagated to veda-routes

ciaransweet:
@botanical I think: `.github/actions/terraform-deploy/action.yml:72` does this? 

Or at least, this issue is probably more relevant to being done in `veda-deploy`?

As the output is defined in `infrastructure/outputs.tf:13`

ciaransweet:
I also don't suppose you have an example of that json file? 😇

ciaransweet:
Closing as it was completed in https://github.com/NASA-IMPACT/veda-deploy/pull/38

PR #196, opened 2024-07-30
feat: update deploy step to output workflows api endpoint

botanical:

Addresses https://github.com/NASA-IMPACT/veda-data-airflow/issues/195

## Changes

* Updated deploy step to output workflows_api endpoint value

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [x] Ad-hoc testing - Deploy changes and test manually (https://github.com/NASA-IMPACT/veda-deploy/actions/runs/10187097538/job/28180375212)
- [ ] Integration tests

### DIFF
--- a/.github/actions/terraform-deploy/action.yml
+++ b/.github/actions/terraform-deploy/action.yml
@@ -70,4 +70,3 @@ runs:
       run: |
         cd ./infrastructure
         terraform output -json > ${HOME}/terraform_outputs.json
-        cat ${HOME}/terraform_outputs.json
--- a/.github/actions/terraform-deploy/action.yml
+++ b/.github/actions/terraform-deploy/action.yml
@@ -68,5 +68,6 @@ runs:
       shell: bash
       working-directory: ${{ inputs.dir }}
       run: |
+        cd ./infrastructure
         terraform output -json > ${HOME}/terraform_outputs.json
         cat ${HOME}/terraform_outputs.json
--- a/.github/actions/terraform-deploy/action.yml
+++ b/.github/actions/terraform-deploy/action.yml
@@ -56,11 +56,16 @@ runs:
       with:
         terraform_version: 1.6.6
 
-    - name: Deploy and output workflows API endpoint
-      id: deploy_and_output_workflows_endpoint
+    - name: Deploy
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/terraform-deploy/action.yml
```diff
@@ -62,3 +62,12 @@ runs:
...
         ./scripts/deploy.sh ${{ inputs.env-file }} <<< deploy
+
+    - name: Output workflows API endpoint
+      id: output_workflows_endpoint
+      shell: bash
+      working-directory: ${{ inputs.dir }}
+      run: |
+        cd ./infrastructure
+        terraform output -json workflows_api > ${HOME}/terraform_outputs.json
+        cat ${HOME}/terraform_outputs.json
```
ciaransweet, line 73: Should this have an `output` defined? 

I know jobs explicitly define an output, not sure what actions do

slesaad, line 73: remove 🐱 ?
also, nit: newline

botanical: Great question! We can define an output like this [example does](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file) in the docs, but it also isn't necessary since we can get the value from the `terraform_outputs.json`. Do you have a preference @ciaransweet ?

botanical: [0ac0061](https://github.com/NASA-IMPACT/veda-data-airflow/pull/196/commits/0ac00619989b53be8b24408deeb3c7660c5f750b)

ciaransweet: I think it depends where you want those output values to be

I dont imagine workspace files are serialised across jobs to be accessible? But explicit outputs can be shared between jobs right?

botanical: Yeah, `terraform_outputs.json` is available to be read within the github workflows that it is run in https://github.com/NASA-IMPACT/veda-deploy/actions/runs/10184498090/job/28173565851 so I think we can forgo adding an `output` and, instead, read from the file directly!

slesaad (approved):

ividito (approved):

PR #201, opened 2024-08-05
Fix Breaking Vector Ignest Changes

ranchodeluxe:
Just some minor bad juju

### DIFF
--- a/docker_tasks/vector_ingest/handler.py
+++ b/docker_tasks/vector_ingest/handler.py
@@ -336,7 +336,7 @@ def alter_datetime_add_indexes_eis(collection: str):
     conn.commit()
 
 
-def handler(event, context):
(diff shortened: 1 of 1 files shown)


docker_tasks/vector_ingest/handler.py
```diff
@@ -381,8 +381,4 @@ def handler(event, context):
 
 
 if __name__ == "__main__":
-    sample_event = {
-        "collection": "eis_fire_newfirepix_2",
-        "href": "s3://covid-eo-data/fireline/newfirepix.fgb",
-    }
-    handler(sample_event, {})
+    handler()
```
smohiudd, line 384: Thanks! this is way better

smohiudd (approved):

PR #212, opened 2024-08-09
Support vector data ingest 

paridhi-parajuli:
**Summary:** Refactor code to support vector data ingest
Addresses [Issue](https://github.com/orgs/US-GHG-Center/projects/1/views/7?pane=issue&itemId=71962132)

## Changes

* Generalized code for vector data ingest and  removed `eis`-related functionalies.
* Change default value for `EXTERNAL_ROLE_ARN` passed to the vector ingest ECS
* Change to `mwaa_stack_conf` for the network configuration of vector ingest ECS
* Add configs for loading data into features database
* Add function to assume AWS IAM role and return temporary credentials

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests


dags/veda_data_pipeline/veda_process_vector_pipeline.py
```diff
@@ -99,8 +99,8 @@
         },
         network_configuration={
             "awsvpcConfiguration": {
-                "securityGroups": vector_ecs_conf.get("VECTOR_SECURITY_GROUP"),
-                "subnets": vector_ecs_conf.get("VECTOR_SUBNETS"),
+                    "securityGroups": mwaa_stack_conf.get("SECURITYGROUPS"),
```
smohiudd, line 102: We are going to continue supporting dual VPCs (Vector + default). Suggest to continue using `VECTOR_SECURITY_GROUP`, `VECTOR_SUBNETS` and `VECTOR_VPC`

cc: @amarouane-ABDELHAK, @ranchodeluxe

smohiudd: This commit should address the issue of optional vector sg and subnets: https://github.com/NASA-IMPACT/veda-data-airflow/commit/bad705e69f872bdd91fa225df2602da9efface73

ranchodeluxe:
Hey @paridhi-parajuli! Sorry about the confusion here. Lots of moving parts

So until we have a flow for users to create their own DAGs (like the CSDA project does and which is far into the future for VEDA) it's expected that certain "clients" might have more custom ingestion rules just like some of the code being removed in `load_to_featuresdb_eis` does for the EIS fire group

What was already in here has a "generic" flow for `load_to_featuresdb` that keys off the s3 bucket prefix being discovered. I know @slesaad didn't want this in GHG code. So if we don't like that pattern then we should probably create a whole new DAG or conditionally force the DAG down a different workflow that just has the "generic" flow split out. 

Does that make sense? But we definitely need most of the code that was removed here back into the system

slesaad:
> Hey @paridhi-parajuli! Sorry about the confusion here. Lots of moving parts
> 
> So until we have a flow for users to create their own DAGs (like the CSDA project does and which is far into the future for VEDA) it's expected that certain "clients" might have more custom ingestion rules just like some of the code being removed in `load_to_featuresdb_eis` does for the EIS fire group
> 
> What was already in here has a "generic" flow for `load_to_featuresdb` that keys off the s3 bucket prefix being discovered. I know @slesaad didn't want this in GHG code. So if we don't like that pattern then we should probably create a whole new DAG or conditionally force the DAG down a different workflow that just has the "generic" flow split out.
> 
> Does that make sense? But we definitely need most of the code for that pathway back

@paridhi-parajuli that's right, maybe I wasn't clear enough with the ask. Create a new dag for the generic stuff, we'll still support the existing dag with the custom logic.

PR #219, opened 2024-08-13
Check length of assume role arns

smohiudd:
In MCP environments we aren't passing assuming role arns so this PR checks if assume_role_arns is empty before passing values to `assume_role_read_arn` and `assume_role_write_arn`

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -96,7 +96,7 @@ resource "local_file" "mwaa_variables" {
       log_group_name          = module.mwaa.log_group_name
       mwaa_execution_role_arn = module.mwaa.mwaa_role_arn
       assume_role_read_arn    = length(var.assume_role_arns) > 0 ? var.assume_role_arns[0] : ""
-      assume_role_write_arn   = length(var.assume_role_arns) > 0 ? var.assume_role_arns[0] : ""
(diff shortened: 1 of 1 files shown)


infrastructure/main.tf
```diff
@@ -95,8 +95,8 @@ resource "local_file" "mwaa_variables" {
       ecs_cluster_name        = module.mwaa.cluster_name
       log_group_name          = module.mwaa.log_group_name
       mwaa_execution_role_arn = module.mwaa.mwaa_role_arn
-      assume_role_read_arn    = var.assume_role_arns[0]
-      assume_role_write_arn   = var.assume_role_arns[0]
+      assume_role_read_arn    = length(var.assume_role_arns) > 0 ? var.assume_role_arns[0] : ""
+      assume_role_write_arn   = length(var.assume_role_arns) > 0 ? var.assume_role_arns[0] : ""
```
ranchodeluxe, line 99: so not `[0]` and `[1]` both `[0]`? just making sure

ranchodeluxe (approved):
just a simple question but looks good

infrastructure/main.tf
smohiudd: we're using the same role for both read and right. But I think I should make it [1]. Let me push a change

Issue #221, opened 2024-08-19
Add Makefile to veda-data-airflow


amarouane-ABDELHAK:
We need to add a Makefile to the veda-data-airflow repository. The purpose of this Makefile is to simplify the deployment and local running of Self-Managed Apache Airflow (SM2A). This will streamline the development process by providing easy-to-use commands for common tasks.

amarouane-ABDELHAK:
PR was created

PR #222, opened 2024-08-19
feat: Adding makefile to data-airflow

amarouane-ABDELHAK:

Addresses https://github.com/NASA-IMPACT/veda-data-airflow/issues/221

## Changes

* Adding Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,75 @@
+SECRET_NAME=""
+ENV_FILE=".env"
+SM2A_FOLDER="sm2a"
+
+CHDIR_SHELL := $(SHELL)
+define chdir
+   $(eval _D=$(firstword $(1) $(@D)))
+   $(info $(MAKE): cd $(_D)) $(eval SHELL = cd $(_D); $(CHDIR_SHELL))
+endef
+
+
+important_message = \
+	@echo "\033[0;31m$(1) \033[0m"
+
+info_message = \
+	@echo "\033[0;32m$(1) \033[0m"
+
+
+count_down = \
+	@echo "Spinning up the system please wait..."; \
+	secs=40 ;\
+	while [ $$secs -gt 0 ]; do \
+		printf "%d\033[0K\r" $$secs; \
+		sleep 1; \
+		: $$((secs--)); \
+	done;
+
+
+.PHONY:
+	clean
+	all
+	test
+
+all: switch-to-sm2a sm2a-local-init sm2a-local-run
+
+test:
+	$(call count_down, 10)
+
+switch-to-sm2a:
+	$(call chdir,${SM2A_FOLDER})
+
+sm2a-local-run: switch-to-sm2a sm2a-local-stop sm2a-cp-dags
+	@echo "Running SM2A"
+	docker compose up -d
+	$(call important_message, "Give the resources a minute to be healthy 💪")
+	$(count_down)
+	$(call info_message, "Please visit http://localhost:8080")
+	echo "username:airflow | password:airflow"
+	echo "To use local SM2A with AWS update ${SM2A_FOLDER}/sm2a-local-config/.env AWS credentials"
+
+sm2a-local-init: switch-to-sm2a sm2a-cp-dags
+	cp sm2a-local-config/env_example sm2a-local-config/.env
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -35,6 +35,31 @@ See [terraform-getting-started](https://developer.hashicorp.com/terraform/tutori
...
 See [getting-started-install](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
 
+
+
+### Setup a local SM2A development environment
+
+1. Initialize the metadata db (only needed once)
+
+```shell
+make sm2a-local-init
```
ciaransweet, line 45: I wonder if this needs to be a make command if it's only ever run once?

Could it just live in the README?

I'm not tied to either approach, but I would assume most make commands would be used frequently

Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+all: switch-to-sm2a sm2a-local-init sm2a-local-run
+
+test:
+	$(call count_down, 10)
+
+switch-to-sm2a:
+	$(call chdir,${SM2A_FOLDER})
+
+sm2a-local-run: switch-to-sm2a sm2a-local-stop sm2a-cp-dags
```
ciaransweet, line 42: I assume you're running `sm2a-local-stop` first to make sure you don't try to run sm2a if it's already running?

Makefile
```diff
@@ -0,0 +1,75 @@
...
+	python scripts/generate_env_file.py --secret-id ${SECRET_NAME} --env-file ${ENV_FILE}
+	./scripts/deploy.sh ${ENV_FILE} <<< init
+	./scripts/deploy.sh ${ENV_FILE} <<< deploy
+
+clean: switch-to-sm2a sm2a-local-stop
+	@echo "Cleaning local env"
+	docker container prune <<< y
+	docker image prune <<< y
+	docker volume prune <<< y
+
```
ciaransweet, line 75:
```suggestion
	docker container prune -f
	docker image prune -f
	docker volume prune -f

```

Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+.PHONY:
+	clean
+	all
+	test
+
+all: switch-to-sm2a sm2a-local-init sm2a-local-run
+
+test:
```
ciaransweet, line 36: I think this name can change... it's not testing anything, it's just counting down seconds, does it even need to be a command? Just keep it as a function in `count_down`?

Makefile
```diff
@@ -0,0 +1,75 @@
...
+	@echo "Spinning up the system please wait..."; \
+	secs=40 ;\
+	while [ $$secs -gt 0 ]; do \
+		printf "%d\033[0K\r" $$secs; \
+		sleep 1; \
+		: $$((secs--)); \
+	done;
+
+
+.PHONY:
```
ciaransweet, line 29: I think all your targets are actually `PHONY` targets.

ciaransweet (changes requested):

README.md
amarouane-ABDELHAK: It was my motivation to have a makefile here, without it, the user would have to type three long command lines instead of simply running 'make init`

Makefile
amarouane-ABDELHAK: Yes, correct

README.md
ciaransweet: I guess I'd assume most would just copy paste 😛

Makefile
amarouane-ABDELHAK: I went with `docker * prune <<< y ` to show the user any warnings or info docker might throwback (can help debugging).

Makefile
amarouane-ABDELHAK: I was using it to test my make targets. But now thinking about, I think it is a good idea to make it run the pytest unittests

ciaransweet (approved):

README.md
amarouane-ABDELHAK: Yes, but I cant count how many time someone missed `cp .env_example .env` while it is stated in README.
. Also, I was mistaken earlier the `init` needs to be run initially to set up the database and create the user. It will need to be run again if `make clean` is executed. I have updated the README to reflect this.

PR #226, opened 2024-08-28
Fix installing the deployment dependencies

amarouane-ABDELHAK:
Fixing the path of the deploy_requirements.txt

Note: It is weird make command did not fail when running the command `pip install deploy_requirements.txt` when the file deploy_requirements.txt does not exist.

### DIFF
--- a/sm2a/Makefile
+++ b/sm2a/Makefile
@@ -1,13 +1,11 @@
 SECRET_NAME=""
 ENV_FILE=".env"
 
-
 important_message = \
 	@echo "\033[0;31m$(1) \033[0m"
 
 info_message = \
 	@echo "\033[0;32m$(1) \033[0m"
-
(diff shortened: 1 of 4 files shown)


anayeaye (approved):

sm2a/Makefile
```diff
@@ -48,7 +48,7 @@ sm2a-local-stop:
 
 sm2a-deploy:
 	@echo "Installing the deployment dependency"
-	pip install -r deploy_requirements.txt
```
anayeaye, line 51: This path looks right though? Is it a cd in the deployment action?

amarouane-ABDELHAK: The command is supposed to run inside `sm2a` folder. deploy_requirements.txt lives in the root folder.

PR #229, opened 2024-09-10
fix/Deploy SM2A via veda-deploy

amarouane-ABDELHAK:

- We can use Veda-deploy to deploy SM2A to different environments
- Added some checks for required environment variables


anayeaye:
we're definitely already doing this :)

PR #244, opened 2024-10-18
Hotfix/eis fire vector support

anayeaye:
## What/why
Recent changes to staging have stopped the current EIS fire data ingest. This PR enables fire vector processing by default and allows generic vector ingests to switch to the generic dag by supplying more info in the payload.


anayeaye:
There was a way to fix this in the event bridge lambda

PR #245, opened 2024-10-18
dev to main BLOCKED by issue #252

smohiudd:
Note: #252 needs to be completed before the next merge into the main branch


infrastructure/variables.tf
```diff
@@ -76,6 +76,11 @@ variable "vector_vpc" {
   default = "null"
 }
 
+variable "deploy_vector_automation" {
+  type    = bool
+  default = "false"
+}
+
```
botanical, line 83: I know this is a PR to promote dev to main, but when it is necessary to write a corresponding change to the variables template file? Do we not need to do that in this case?

infrastructure/s3_event_bridge_lambda.tf
```diff
@@ -0,0 +1,203 @@
+#####################################################
+# Variables
+#####################################################
+variable "DEPLOY_VECTOR_AUTOMATION" {
+  type    = bool
+  default = false
+}
```
botanical: do we need this if it's defined in the variables file? also, it's uppercase 😯 I think the idiomatic style is lowercase

infrastructure/functions/s3_event_bridge_to_sfn_execute/lambda_function.py
```diff
@@ -0,0 +1,60 @@
+import boto3
+import http.client
+import os
+import base64
+import ast
+import json
+
+
+#mwaa_env_name = 'veda-pipeline-dev-mwaa'
```
botanical:
```suggestion
```

infrastructure/s3_event_bridge_lambda.tf
smohiudd: good catch! We don't need this since it's already in `variables.tf`. I'll delete it

infrastructure/variables.tf
smohiudd: yes we will need TF_VAR_deploy_vector_automation=True for staging

infrastructure/s3_event_bridge_lambda.tf
```diff
@@ -0,0 +1,203 @@
...
+#####################################################
+resource "aws_lambda_permission" "s3_invoke" {
+  count = var.deploy_vector_automation ? 1 : 0
+
+  provider = aws.aws_current
+  action           = "lambda:InvokeFunction"
+  function_name    = aws_lambda_function.lambda[0].function_name
+  principal        = "s3.amazonaws.com"
+  statement_id     = "AllowInvocationFromS3Bucket-veda-${var.stage}"
+  source_account   = "114506680961"
```
botanical: Sorry if you explained this already but why do we need to hard code this value?

botanical (approved):
Had a couple questions but this looks good

PR #261, opened 2024-11-25
Fixing collecton ingest DAG

amarouane-ABDELHAK:

Fixing collection ingest 
## Changes

* Ingest Collection DAG

### DIFF
--- a/dags/veda_data_pipeline/groups/collection_group.py
+++ b/dags/veda_data_pipeline/groups/collection_group.py
@@ -1,11 +1,9 @@
 import requests
 from airflow.models.variable import Variable
-from airflow.operators.python import PythonOperator
 from airflow.utils.task_group import TaskGroup
+from airflow.decorators import task
 from veda_data_pipeline.utils.collection_generation import GenerateCollection
 from veda_data_pipeline.utils.submit_stac import submission_handler
-
-generator = GenerateCollection()
 
 
 def check_collection_exists(endpoint: str, collection_id: str):
@@ -24,27 +22,7 @@ def check_collection_exists(endpoint: str, collection_id: str):
     )
 
 
-def ingest_collection_task(ti):
-    """
-    Ingest a collection into the STAC catalog
 
-    Args:
-        dataset (Dict[str, Any]): dataset dictionary (JSON)
-        role_arn (str): role arn for Zarr collection generation
-    """
-    import json
-    collection = ti.xcom_pull(task_ids='Collection.generate_collection')
-    airflow_vars = Variable.get("aws_dags_variables")
-    airflow_vars_json = json.loads(airflow_vars)
-    cognito_app_secret = airflow_vars_json.get("COGNITO_APP_SECRET")
-    stac_ingestor_api_url = airflow_vars_json.get("STAC_INGESTOR_API_URL")
-
-    return submission_handler(
-        event=collection,
-        endpoint="/collections",
-        cognito_app_secret=cognito_app_secret,
-        stac_ingestor_api_url=stac_ingestor_api_url
-    )
 
 
 # NOTE unused, but useful for item ingests, since collections are a dependency for items
@@ -60,19 +38,6 @@ def check_collection_exists_task(ti):
     )
 
 
-def generate_collection_task(ti):
-    import json
-    config = ti.dag_run.conf
-    airflow_vars = Variable.get("aws_dags_variables")
-    airflow_vars_json = json.loads(airflow_vars)
-    role_arn = airflow_vars_json.get("ASSUME_ROLE_READ_ARN")
-
-    # TODO it would be ideal if this also works with complete collections where provided - this would make the collection ingest more re-usable
-    collection = generator.generate_stac(
-        dataset_config=config, role_arn=role_arn
-    )
-    return collection
-
 
 
 group_kwgs = {"group_id": "Collection", "tooltip": "Collection"}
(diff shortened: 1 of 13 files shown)


dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -45,57 +43,21 @@
...
 
-    build_stac = build_stac_task.expand(payload=get_files)
-    # .output is needed coming from a non-taskflow operator
-    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac)
+    run_discover_build_and_push = TriggerMultiDagRunOperator(
+        task_id="trigger_discover_items_dag",
+        dag=dag,
+        trigger_dag_id="veda_discover",
+        python_callable=trigger_discover_and_build_task,
+    )
```
ividito: We should not use this operator, as it creates a disconnect between the original DAGRun event and subsequent DAG runs (ie. a failure in an instance veda-discover will not feed back to the original veda-dataset DAG using this operator). We could instead describe the discover pipeline in a TaskGroup and call `expand()` on it, which will map task instances to the correct event.

ividito (changes requested):
Is there additional context to this PR? I didn't know we had issues with collection ingests.

dags/veda_data_pipeline/veda_dataset_pipeline.py
amarouane-ABDELHAK: I want to re-use `discover_build_ingest` DAG instead of redefining the same tasks. But having a task group used in both DAGs is not a bad idea

sm2a/airflow_worker/Dockerfile
```diff
@@ -41,7 +41,9 @@ COPY --chown=airflow:airflow scripts "${AIRFLOW_HOME}/scripts"
 
 RUN cp ${AIRFLOW_HOME}/configuration/airflow.cfg* ${AIRFLOW_HOME}/.
 
-RUN pip install pypgstac==0.7.4
+# Commited because it downgrade pydentics to v1
```
anayeaye: This is an artifact and we should not need pypgstac in airflow--that should be entirely handled by the ingest api

dags/veda_data_pipeline/groups/collection_group.py
```diff
@@ -60,32 +38,48 @@ def check_collection_exists_task(ti):
...
+            airflow_vars_json = Variable.get("aws_dags_variables", deserialize_json=True)
+            cognito_app_secret = airflow_vars_json.get("COGNITO_APP_SECRET")
+            stac_ingestor_api_url = airflow_vars_json.get("STAC_INGESTOR_API_URL")
+
+            return submission_handler(
+                event=collection,
+                endpoint="/collections",
+                cognito_app_secret=cognito_app_secret,
+                stac_ingestor_api_url=stac_ingestor_api_url
+            )
```
ividito, line 80: Let's define these `task()` functions outside of the TaskGroup, so they can be re-used on their own as well.

dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -44,58 +43,21 @@
...
-    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac)
+    items = start >> collection_task_group() >> get_items()
+    run_discover_build_and_push = TriggerDagRunOperator.partial(
+        task_id="trigger_discover_items_dag",
+        trigger_dag_id="veda_discover",
+        wait_for_completion=True,
 
-    collection_grp.set_upstream(start)
-    submit_stac.set_downstream(end)
+    ).expand(conf=items) >> end
```
ividito: It looks like this will allow the parent DAG to properly track the state of mapped+triggered DAGs, but still recreates a disconnect (the link created does not associate DAG runs with a parent DAG, it only associates the DAG itself with its parent).

ividito (changes requested):

dags/veda_data_pipeline/groups/discover_group.py
```diff
@@ -48,6 +49,36 @@ def discover_from_s3_task(ti=None, event={}, **kwargs):
 
 
 @task
+def get_files_task(payload, ti=None):
+    """
+    Get files from S3 produced by discovery or dataset tasks.
+    Handles both single payload and multiple payload scenarios.
+    """
```
ividito, line 56: 🚀, this is much nicer than what we were doing before

dags/veda_data_pipeline/veda_discover_pipeline.py
```diff
@@ -98,7 +98,7 @@ def get_discover_dag(id, event=None):
         # define DAG using taskflow notation
 
         discover = discover_from_s3_task(event=event)
-        get_files = get_files_to_process(payload=discover)
+        get_files = get_files_task(payload=discover)
         build_stac = build_stac_task.expand(payload=get_files)
         # .output is needed coming from a non-taskflow operator
```
ividito: Still true, but no longer applicable
```suggestion

```

dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -83,19 +64,38 @@ def build_stac_task(payload):
...
+                payload.pop("assets")
+            # finally put the mutated assets back in the payload
+            else:
+                payload["assets"] = assets
+        for item in payload.get("discovery_items"):
+            payloads.append({
+                **payload,
+                **item
+            }
+            )
```
ividito, line 91: This adds a lot of additional unknown properties to the payload, which will cause downstream issues (especially as we add enforced schemas and `Param` definitions)

dags/veda_data_pipeline/groups/collection_group.py
ividito: ^^ this is a breaking change for `stactools`, which come with their own collection generators, but re-use our collection submission task

dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -83,19 +64,38 @@ def build_stac_task(payload):
...
-    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac)
 
-    collection_grp.set_upstream(start)
-    submit_stac.set_downstream(end)
+    mutated_payloads = start >> collection_task_group() >> mutate_payload()
+    run_discover_build_and_push = TriggerDagRunOperator.partial(
+        task_id="trigger_discover_items_dag",
+        trigger_dag_id="veda_discover",
+        wait_for_completion=True
+    ).expand(conf=mutated_payloads) >> end
```
ividito, line 101: I get that this promotes reuseability at the level of the `discover` DAG, but I worry that this creates too much of a breakdown in reuse and observability at the task level.

- Rather than passing task-specific data structures as parameters, we end up relying on `ti.dag_run.conf`. This means that, unless we always build and ingest items through the discover DAG, we cannot rely on the data format being the same for different DAGs with different ingestion strategies. This will eventually bring us back to the same solution we have now, where new DAGs skip using the `discover` DAG, and simply reuse the tasks with additional wrapper steps to manipulate the input. We should try to lean into this - the changes you made to `get_files()` is a great example, where we promote modularity and reuse at the task level, so that it doesn't matter what the incoming `dag_run.conf` is.
- Similarly, this condenses task status into a single node in the `dataset` DAG. Meanwhile, in the triggered `discover` DAG, there are several expanded steps with unique failure conditions. This means that a failure in one step in one triggered DAG will require a retry of the complete DAG, rather than a single task.
- `TriggerDagRunOperator` adds a link between DAGs, but not between executions. I was hoping this would work better with `expand()`, but unfortunately we're out of luck. This (and a few similar issues) is being [tracked as a bug ](https://github.com/apache/airflow/issues/41977) in Airflow, but until it's fixed, I think we should steer clear so as to maintain execution-level observability, especially in DAG runs with a large number of mapped discovery tasks.

ividito (changes requested):

dags/veda_data_pipeline/veda_dataset_pipeline.py
amarouane-ABDELHAK: Yes, I agree that traceability will be necessary for debugging any issues. Some have reported that this problem is fixed in Airflow 2.10. However, we still need to promote task reusability to avoid defining the same logic twice. I will close this PR and open a new one with some refactoring to make debugging and development easier.

amarouane-ABDELHAK: Closing this in favor of https://github.com/NASA-IMPACT/veda-data-airflow/pull/268

PR #267, opened 2024-11-27
Strip thumbnail assets from payload before discovery

ividito:
Mutate payload before discovery to ignore thumbnail assets meant for the collection level.

### DIFF
--- a/dags/veda_data_pipeline/veda_dataset_pipeline.py
+++ b/dags/veda_data_pipeline/veda_dataset_pipeline.py
@@ -61,6 +61,18 @@ def build_stac_task(payload):
     event_bucket = airflow_vars_json.get("EVENT_BUCKET")
     return stac_handler(payload_src=payload, bucket_output=event_bucket)
 
+@task()
+def mutate_payload(**kwargs):
+    ti = kwargs.get("ti")
+    payload = ti.dag_run.conf
+    if assets := payload.get("assets"):
+        # is first key thumbnail
+        if "thumbnail" in assets.keys():
+            assets.pop("thumbnail")
+        if not assets:
+            payload.pop("assets")
+    return payload
+
 
 template_dag_run_conf = {
     "collection": "<collection-id>",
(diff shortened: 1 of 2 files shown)


dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -61,6 +61,18 @@ def build_stac_task(payload):
...
+def mutate_payload(**kwargs):
+    ti = kwargs.get("ti")
+    payload = ti.dag_run.conf
+    if assets := payload.get("assets"):
+        # is first key thumbnail
+        if "thumbnail" in assets.keys():
+            assets.pop("thumbnail")
+        if not assets:
+            payload.pop("assets")
+    return payload
```
anayeaye:
```suggestion
def mutate_payload(**kwargs):
    ti = kwargs.get("ti")
    payload = ti.dag_run.conf
    if assets := payload.get("assets"):
        # remove thumbnail asset if provided in collection config
        if "thumbnail" in assets.keys():
            assets.pop("thumbnail")
        # if thumbnail was only asset, delete assets
        if not assets:
            payload.pop("assets")
        # finally put the mutated assets back in the payload
        else:
            payload["assets"] = assets
    return payload
```

dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -98,4 +115,5 @@ def build_stac_task(payload):
     submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac)
 
     collection_grp.set_upstream(start)
+    mutate_payload_task.set_upstream(start)
```
anayeaye, line 118: should `collection_grp` aslo be upstream of mutate_payload_task so we don't strip assets before the collection is created?

ividito: Nope - it's not an in-place mutation, the output gets passed to the discover step independantly.

anayeaye: Resolving--the discover task, which includes the `alt_payload=mutate_payload_task()` parameter, depends on `collection_grp` already

anayeaye (approved):

PR #275, opened 2025-01-15
fix(transfer): add non-decorated shared transfer function

ividito:
To avoid ValueErrors when calling transfer_data, refactor tasks to invoke a shared transfer function, rather than calling a task from another task (which caused issues with the `ti` parameter).

### DIFF
--- a/dags/veda_data_pipeline/groups/transfer_group.py
+++ b/dags/veda_data_pipeline/groups/transfer_group.py
@@ -27,11 +27,8 @@ def cogify_copy_task(ti):
     return cogify_transfer_handler(event_src=config, external_role_arn=external_role_arn)
 
 @task
-def transfer_data(ti, payload):
+def transfer_data_task(ti=None, payload={}):
     """Transfer data from one S3 bucket to another; s3 copy, no need for docker"""
-    from veda_data_pipeline.utils.transfer import (
-        data_transfer_handler,
-    )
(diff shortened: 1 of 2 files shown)


dags/veda_data_pipeline/groups/transfer_group.py
```diff
@@ -42,12 +39,21 @@ def transfer_data(ti, payload):
...
+    transfer_data(config)
+
+# non-decorated function for use in other tasks
+def transfer_data(payload={}):
+    """Transfer data from one S3 bucket to another; s3 copy, no need for docker"""
+    from veda_data_pipeline.utils.transfer import (
+        data_transfer_handler,
+    )
+    # use task-provided payload if provided, otherwise fall back on ti values
+    # payload will generally have the same values expected by discovery, so some renames are needed when combining the dicts
```
ividito:
```suggestion
```

anayeaye (approved):
The payload transformation is running in dev sm2a and seems to be working as expected. We do see a data access permission issue from the ingest API which is unexpected but also the dev api is not intended to generate metadata for assets in production anyway.

For the promotion step I think we may want to ask users (and automated events) to add both
`transfer: true` and
`target_bucket: veda-data-store` or the instance production bucket equivalent in configuration for the promotion DAG. In this PR all defaults are for veda production. What do you think?
cc: @smohiudd @slesaad @botanical

Issue #287, opened 2025-01-24
Resolve Key error in veda_dataset_pipeline


anayeaye:
## What
Small task: veda_dataset_pipeline currently removes assets object when mutating payload between tasks but should not fail when no assets object is provided in dataset config.

```
[2025-01-24, 18:41:33 UTC] {taskinstance.py:2731} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 444, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 414, in _execute_callable
    return execute_callable(context=context, **execute_callable_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/decorators/base.py", line 241, in execute
    return_value = super().execute(context)
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/operators/python.py", line 200, in execute
    return_value = self.execute_callable()
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/operators/python.py", line 217, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/dags/veda_data_pipeline/groups/processing_tasks.py", line 29, in remove_thumbnail_asset
    payload.pop("assets")
KeyError: 'assets'
```

## AC
- [ ] veda_dataset_pipeline does not fail when optional assets key not provided in dataset config

smohiudd:
Options: 
- fix discovery objects nested assets (tech debt)
- enforce schema - can't create dataset config without thumbnail asset

[...]

PR #293, opened 2025-01-29
fix: Updates to fail gracefully if there are no assets to resolve key_error

botanical:
Addresses [VEDA-287: Resolve Key error in veda_dataset_pipeline](https://github.com/NASA-IMPACT/veda-data-airflow/issues/287)

## Changes

* Updates to fail gracefully if there are no assets for  `payload.pop("assets")`

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/dags/veda_data_pipeline/groups/processing_tasks.py
+++ b/dags/veda_data_pipeline/groups/processing_tasks.py
@@ -25,8 +25,9 @@ def remove_thumbnail_asset(ti):
     if assets.get("thumbnail"):
         assets.pop("thumbnail")
     # if thumbnail was only asset, delete assets
-    if not assets and (assets is not None):
-        payload.pop("assets")
+    if not assets:
+        # Safely return if there are no assets in the payload
+        payload.pop("assets", True)
     return payload
 
 # with exponential backoff enabled, retry delay is converted to seconds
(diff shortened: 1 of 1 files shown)


dags/veda_data_pipeline/groups/processing_tasks.py
```diff
@@ -25,7 +25,7 @@ def remove_thumbnail_asset(ti):
     if assets.get("thumbnail"):
         assets.pop("thumbnail")
     # if thumbnail was only asset, delete assets
-    if not assets:
+    if not assets and (assets is not None):
```
anayeaye: ~We need to update line 24 too, so that we can test if assets exists~

~`assets = payload.get("assets", {})` should be `assets = payload.get("assets", None)`~

EDIT
I think having an assets key at all is a problem for the existing pipeline because the pipeline and stac schemas have conflicting `assets` meanings--so this level of logic about other non-thumbnail assets seems unhelpful. @ividito @botanical should we just check if assets is in payload.keys() and then if it is payload.pop("assets")?

smohiudd: I agree that the current logic is awkward. I also think we should just pop the assets from the payload.

anayeaye (approved):

smohiudd (approved):

ividito (approved):

Issue #288, opened 2025-01-24
Use STAC transactions instead of Ingest API in Submit STAC Util


smohiudd:
## Description

The [submit_stac](https://github.com/NASA-IMPACT/veda-data-airflow/blob/dev/dags/veda_data_pipeline/utils/submit_stac.py) data ingestion util uses the veda-backend /api/ingest/ingestions API to load data into the STAC database. The Ingest API also includes a DynamoDB to queue ingestion. Because of this setup, it is difficult the ingest API does not currently propagate errors to DAGs in airflow and there it's not currently possible to get error feedback on ingestions. This is a particular problem for ingestion automation since we can can pass errors to Github workflows or the veda-ingest-ui.

We can instead use the STAC API transactions [bulk ingestions](http://dev.openveda.cloud/api/stac/docs#/Bulk%20Transaction%20Extension/Bulk_Create_Item_collections__collection_id__bulk_items_post) or single STAC item to load STAC items and are able to get error feedback immediately in the submit stac util.

## Acceptance Criteria

- [ ] Use STAC API bulk transactions endpoint in submit stac util

smohiudd:
- MAAP is not currently using transactions API, still using the ingest API. May plan to deploy transactions if its included in eoAPI.
- @jjfrench to get feedback on eoAPI roadmap for transactions support
- will be difficult to maintain support for ingest API and transactions concurrently in the same DAG. MAAP can write a separate DAG using ingest API.

smohiudd:
Testing plan:
- deploy to SIT
- ensure feature flag is working
- ingest using transactions API
- check  if summaries, spatial extent updates (uses ingest API)

PR #297, opened 2025-02-13
Submit STAC using transactions API

ividito:
Proposed implementation for #288 

Original ingest API support is maintained through an Airflow variable. I learned while setting this up that this is [not a recommended approach](https://airflow.apache.org/docs/apache-airflow/stable/howto/dynamic-dag-generation.html#dynamic-dags-with-environment-variables), but it is consistent with what we do elsewhere.

### DIFF
--- a/dags/veda_data_pipeline/groups/processing_tasks.py
+++ b/dags/veda_data_pipeline/groups/processing_tasks.py
@@ -51,7 +51,6 @@ def submit_to_stac_ingestor_task(built_stac: dict):
     """Submit STAC items to the STAC ingestor API."""
     event = built_stac.copy()
     success_file = event["payload"]["success_event_key"]
-
     try:
         success_file = event["payload"]["success_event_key"]
         with smart_open.open(success_file, "r") as _file:
--- a/dags/veda_data_pipeline/groups/processing_tasks.py
+++ b/dags/veda_data_pipeline/groups/processing_tasks.py
@@ -37,14 +37,12 @@ if TRANSACTIONS_ENDPOINT_ENABLED:
     # assuming default chunk size (500), this matches the current dynamoDB configuration on the STAC ingestor
     task_kwargs = {"retries": 3, "retry_delay": 10, "retry_exponential_backoff": True, "max_active_tis_per_dag": 2}
     submit_kwargs = {}
-    submit_handler = submit_transactions_handler
-    ingest_url = airflow_vars_json.get("STAC_INGESTOR_API_URL")
+    ingest_url = airflow_vars_json.get("STAC_URL")
     app_secret = airflow_vars_json.get("STAC_API_KEYCLOAK_CLIENT_SECRET")
 else:
     task_kwargs = {"retries": 2, "retry_delay": 60, "retry_exponential_backoff": True, "max_active_tis_per_dag": 5}
     submit_kwargs = {"endpoint": "/ingestions"}
-    submit_handler = submission_handler
-    ingest_url = airflow_vars_json.get("STAC_URL")
+    ingest_url = airflow_vars_json.get("STAC_INGESTOR_API_URL")
     app_secret = airflow_vars_json.get("COGNITO_APP_SECRET")
 
 # with exponential backoff enabled, retry delay is converted to seconds
@@ -62,13 +60,21 @@ def submit_to_stac_ingestor_task(built_stac: dict):
         log_task("No success file found - using event directly")
         stac_items = [event]
 
-    for item in stac_items:
-        submit_handler(
-            event=item,
+    if TRANSACTIONS_ENDPOINT_ENABLED:
+        submit_transactions_handler(
+            event=stac_items,
             cognito_app_secret=app_secret,
             ingest_url=ingest_url,
             **submit_kwargs,
         )
+    else:
+        for item in stac_items:
+            submission_handler(
+                event=item,
+                cognito_app_secret=app_secret,
+                ingest_url=ingest_url,
+                **submit_kwargs,
+            )
     return event
 
 @task(retries=2, retry_delay=60, retry_exponential_backoff=True, max_active_tis_per_dag=5)
--- a/dags/veda_data_pipeline/groups/processing_tasks.py
+++ b/dags/veda_data_pipeline/groups/processing_tasks.py
@@ -39,11 +39,13 @@ if TRANSACTIONS_ENDPOINT_ENABLED:
     submit_kwargs = {}
     submit_handler = submit_transactions_handler
     ingest_url = airflow_vars_json.get("STAC_INGESTOR_API_URL")
+    app_secret = airflow_vars_json.get("STAC_API_KEYCLOAK_CLIENT_SECRET")
 else:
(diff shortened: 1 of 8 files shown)


dags/veda_data_pipeline/utils/submit_stac_transactions.py
```diff
@@ -0,0 +1,121 @@
...
+class AppConfig(TypedDict):
+    cognito_domain: str
+    client_id: str
+    client_secret: str
+    scope: str
+
+class TransactionsApi:
+
+    @classmethod
+    def from_veda_auth_secret(cls, *, secret_id: str, base_url: str) -> "IngestionApi":
```
amarouane-ABDELHAK: Shouldn't the return type hint be "TransactionsApi" instead of "IngestionApi"

dags/veda_data_pipeline/utils/submit_stac_transactions.py
```diff
@@ -0,0 +1,121 @@
...
+                "grant_type": "client_credentials",
+                # A space-separated list of scopes to request for the generated access token.
+                "scope": scope,
+            },
+        )
+        try:
+            response.raise_for_status()
+        except Exception as ex:
+            print(response.text)
+            raise f"Error, {ex}"
```
amarouane-ABDELHAK: Raise should be derived from a baseException otherwise it won't raise a string 🤔
Probably we should use
```python
raise RuntimeError(f"Error, {ex}")

```

dags/veda_data_pipeline/utils/submit_stac_transactions.py
```diff
@@ -0,0 +1,121 @@
...
+        return response.json()
+
+    def __init__(self, stac_ingestor_api_url: str, cognito_app_secret: str = None):
+        """
+        :param stac_endpoint: Base URL of the STAC API (e.g., 'https://example.com/stac').
+        :param token: Optional Bearer token for authenticated STAC APIs.
+        """
+        self.stac_ingestor_api_url = stac_ingestor_api_url.rstrip('/')
+        self.cognito_app_secret = cognito_app_secret
```
amarouane-ABDELHAK: Where are we using `self.cognito_app_secret`?

amarouane-ABDELHAK (commented):
Thank you for the PR. I just left some small comments

amarouane-ABDELHAK (approved):

smohiudd:
This branch has been deployed to SIT, but testing is currently blocked because the dev STAC API has switched over to using veda-keycloak and we still need to sort out how programatic auth will work.

smohiudd:
This branch was tested by switching on and off the `TF_VAR_transactions_endpoint_enabled` feature flag. I ingested the `ida-ndwi-difference-test-transactions` collection into the [dev catalog](https://dev.openveda.cloud/stac/collections/ida-ndwi-difference-test-transactions).

sm2a/infrastructure/main.tf
```diff
@@ -104,7 +104,9 @@ module "sma-base" {
     ASSUME_ROLE_WRITE_ARN = var.assume_role_write_arn,
     SM2A_BASE_URL         = module.sma-base.airflow_url,
     CLOUDFRONT_TO_INVALIDATE = var.cloudfront_to_invalidate
-    CLOUDFRONT_PATH_TO_INVALIDATE = var.cloudfront_path_to_invalidate
+    CLOUDFRONT_PATH_TO_INVALIDATE = var.cloudfront_path_to_invalidate,
+    TRANSACTIONS_ENDPOINT_ENABLED = var.transactions_endpoint_enabled==true ? "True" : null
```
smohiudd: I don't think tf bools are serialized in airflow in python. I used this hack but open if there's a better way.

dags/veda_data_pipeline/groups/processing_tasks.py
```diff
@@ -1,14 +1,17 @@
...
 from copy import deepcopy
 import smart_open
 from airflow.models.variable import Variable
 from airflow.decorators import task
 from veda_data_pipeline.utils.submit_stac import submission_handler
+from veda_data_pipeline.utils.submit_stac_transactions import submit_transactions_handler
 
 group_kwgs = {"group_id": "Process", "tooltip": "Process"}
 
+airflow_vars = Variable.get("aws_dags_variables")
```
botanical, line 12: If we set deserialize_json=True, do you know if we still need to do:
```
var.transactions_endpoint_enabled==true ? "True" : null
```
in main.tf? 

```suggestion
airflow_vars = Variable.get("aws_dags_variables", deserialize_json=True)
```

And maybe we refactor this file a bit so we call Variable.get in tasks rather than in the top level, since the docs seem to imply this is best practice: https://airflow.apache.org/docs/apache-airflow/2.8.4/best-practices.html#airflow-variables

sm2a/infrastructure/main.tf
```diff
@@ -103,8 +103,10 @@ module "sma-base" {
     ASSUME_ROLE_READ_ARN  = var.assume_role_read_arn,
     ASSUME_ROLE_WRITE_ARN = var.assume_role_write_arn,
     SM2A_BASE_URL         = module.sma-base.airflow_url,
-    CLOUDFRONT_TO_INVALIDATE = var.cloudfront_to_invalidate
-    CLOUDFRONT_PATH_TO_INVALIDATE = var.cloudfront_path_to_invalidate
+    CLOUDFRONT_TO_INVALIDATE = var.cloudfront_to_invalidate,
+    CLOUDFRONT_PATH_TO_INVALIDATE = var.cloudfront_path_to_invalidate,
+    TRANSACTIONS_ENDPOINT_ENABLED = var.transactions_endpoint_enabled==true ? "True" : null,
```
ividito, line 108: Related fix that removes the need for the ternary here: https://github.com/NASA-IMPACT/self-managed-apache-airflow/pull/18

cc/ @amarouane-ABDELHAK

dags/veda_data_pipeline/groups/processing_tasks.py
ividito: I would love a refactor for this (we call the same set of `Variable.get("aws_dags_variables")` lines 16 times in our DAGs), but I think that's a different issue. `.airflowignore` keeps the DAG processor from parsing the `groups` and `utils` folders, so there's no performance loss by having these calls here.

smohiudd: I tried using `airflow_vars_json = Variable.get("aws_dags_variables", deserialize_json=True)` and I'm getting a decode error: `requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

ividito:
^^ rebased to pull in changes from dev

smohiudd:
Based on conversations with stac-fastapi maintainers, they recommend we don't use bulk ingestion. I tried instead to use the PUT {collection_id}/items/{item_id} endpoint for create and update operations but stac-fastapi only supports [updates](https://github.com/stac-utils/stac-fastapi/blob/main/stac_fastapi/extensions/stac_fastapi/extensions/core/transaction.py#L107). 

The submit stac util will need to be refactored to handle both creating new items via POST and updating items via PUT. 

@ividito I suggest we proceed with mwaa refactor PR since this transactions PR will need more time.

PR #298, opened 2025-02-19
Update vector ingestion

paridhi-parajuli:
There are two dags `veda-ingest-vector` and `veda-generic-ingest-vector` that have the same `handler.py`.
This PR removes the redundant `veda-generic-ingest-vector` and performs some changes to the existing `veda-ingest-vector` dag.
## Changes:

- [x] Modify the dag config.

For Non EIS data:
  - [x] Update the code to pass and take `extra flags` when provided
  - [x] Modify the code to take filename and `id_template` when collection is not provided in the dag conf


amarouane-ABDELHAK (approved):

smohiudd:
Can we put a hold on this for now. We need to discus the impacts to EIS fire ingestion...

cc: @anayeaye

smohiudd:
Based on these comments from last year I thought we were going to maintain two separate DAGs: one for generic and one EIS specific: https://github.com/NASA-IMPACT/veda-data-airflow/pull/212#issuecomment-2278296607

But I can see that the EIS fire handler already handled [both cases](https://github.com/NASA-IMPACT/veda-data-airflow/blob/2af4a7cc0f5fc8b30805a2c8a638cbf5fcf7ea3f/docker_tasks/vector_ingest/handler.py#L366-L369): 
```
        if s3_object_prefix.startswith("EIS/"):
            coll_status = load_to_featuresdb_eis(downloaded_filepath, collection)
        else:
            coll_status = load_to_featuresdb(downloaded_filepath, collection)
```

@paridhi-parajuli I wanted to confirm that the original EIS handler is already "generic" enough that it can handle non EIS datasets?

paridhi-parajuli:
@smohiudd The changes I have made so far doesnt interfere with any of the EIS related functionality. I am not sure how veda is planning to handle non-EIS vector datasets, but for ghgc use case so far this should generic enough.

smohiudd:
Thanks @paridhi-parajuli for clarifying. I'm ok with this PR. We should be ok to merge!

Issue #295, opened 2025-02-11
Standalone transfer DAG in SM2A fails with not VALID_BUCKET error


anayeaye:
## What 
When called directly from SM2A, the transfer DAG fails with error about invalid bucket. This may be related to recent changes for the veda_promotion_pipeline which uses the transfer DAG.

**trace**
```
Traceback (most recent call last):
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 444, in _execute_task
    result = _execute_callable(context=context, **execute_callable_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/models/taskinstance.py", line 414, in _execute_callable
    return execute_callable(context=context, **execute_callable_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/operators/python.py", line 200, in execute
    return_value = self.execute_callable()
                   ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/airflow/miniconda3/envs/py11/lib/python3.11/site-packages/airflow/operators/python.py", line 217, in execute_callable
    return self.python_callable(*self.op_args, **self.op_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/dags/veda_data_pipeline/groups/transfer_group.py", line 54, in transfer_data
    return data_transfer_handler(event=payload, role_arn=external_role_arn)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/dags/veda_data_pipeline/utils/transfer.py", line 97, in data_transfer_handler
    matching_files = get_matching_files(
                     ^^^^^^^^^^^^^^^^^^^
  File "/opt/airflow/dags/veda_data_pipeline/utils/transfer.py", line 25, in get_matching_files
[...]

PR #301, opened 2025-02-25
fix: use configuration dict passed into dag run for transfer_data

botanical:

Addresses [VEDA-295: Develop amazing new feature](https://github.com/NASA-IMPACT/veda-data-airflow/issues/295)

## Changes

* Updates `cogify_branching` to explicitly check for `true` or `True` because, when testing with `cogify: 'false'`, it would cofigy and copy data.
* Updates `transfer_data` to rely on `ti.dag_run.conf` if payload is {}. 

## PR Checklist

- [ ] Update CHANGELOG
- [ ] Unit tests
- [ ] Integration tests

### DIFF
--- a/dags/veda_data_pipeline/groups/transfer_group.py
+++ b/dags/veda_data_pipeline/groups/transfer_group.py
@@ -9,10 +9,10 @@ group_kwgs = {"group_id": "Transfer", "tooltip": "Transfer"}
 
 
 def cogify_choice(ti):
-    """Choos whether to cogify or not; if yes, use a docker container"""
+    """Choose whether to cogify or not; if yes, use a docker container"""
     payload = ti.dag_run.conf
 
-    if payload.get("cogify"):
+    if payload.get("cogify") == "true" or payload.get("cogify") == True:
         return f"{group_kwgs['group_id']}.cogify_and_copy_data"
     else:
         return f"{group_kwgs['group_id']}.copy_data"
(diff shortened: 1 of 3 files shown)


dags/veda_data_pipeline/groups/transfer_group.py
```diff
@@ -51,6 +51,8 @@ def transfer_data(payload={}):
     airflow_vars_json = json.loads(airflow_vars)
     external_role_arn = airflow_vars_json.get("ASSUME_ROLE_WRITE_ARN")
     # (event, chunk_size=2800, role_arn=None, bucket_output=None):
+    if payload == {}:
+        payload = ti.dag_run.conf
```
botanical, line 55: This kind of feels like a code smell 🤧 I would love feedback on this if it is!

anayeaye (approved):

dags/veda_data_pipeline/groups/transfer_group.py
anayeaye: Yeah the payload vs run configuration management feels odd but this change looks like it will fix the standalone transfer for now without causing problems for the in-promotion-pipeline task.

anayeaye (approved):

ividito (approved):

dags/veda_data_pipeline/veda_transfer_pipeline.py
```diff
@@ -37,7 +38,7 @@
     "filename_regex": "<file_regex>",
     "target_bucket": "<target_bucket>",
     "collection": "<collection-id>",
-    "cogify": "true|false",
+    "cogify": Param(type="boolean"),
```
amarouane-ABDELHAK: Can we give cogify an explicit default value 
```python
"cogify": Param(default=False, type="boolean")
```
I know this is out of scope, but can we add an explicit type for dry_run as well?

botanical: https://github.com/NASA-IMPACT/veda-data-airflow/pull/301/commits/cd62aa92a41e7b18baf5886aee2613d946c8c984

Issue #277, opened 2025-01-16
STAC Tools DAG Testing


smohiudd:
## Description

A [STAC Tools demo](https://github.com/NASA-IMPACT/veda-data-airflow/tree/demo/stactools) was developed to work with stac tool packages for data transformation and and STAC creation. There have been major changes to the veda-data-airflow repo since this work started so the latest SM2A changes will need to be merged into this repo. 

Before STAC Tool DAGs are developed and incorporated into `dev` SM2A further testing and validation is required. In particular we need to understand how stactools packages can be passed as configuration to a DAG.

## Acceptance Criteria

- [ ] Test STAC Tools DAG in SM2A dev with example stactools packages.

PR #307, opened 2025-03-05
feat: Add stactool DAGs

ividito:
### **Summary:**

Addresses #277 

### **stactools-pipeline DAGs**

* Adds generic/dynamic pipeline, as well as a `noaa-hrrr` pipeline to illustrate how more customized stactools packages can be implemented
* The generic pipeline has been tested on `sentinel2`, `cop-dem`, and `landsat` datasets with varied results.
* `sentinel2` doesn't implement `create_collection()`, and so it fails with an `AttributeError`
* `landsat` requires a metadata file href as a granule, and works without additional parameters (although I have questions about how it will handle s3-hosted data)
* `cop_dem` works with additional parameters (this test case is included as the default input for the DAG)
* The `noaa-hrrr` package requires unique inputs, and would not work when used in the generic pipeline. 

### How to test

- Both DAGs have been pre-loaded with working, valid inputs. Run sm2a locally, point it at a working ingest API (either local or otherwise - I used local to avoid setting up auth), and trigger the DAGs with the default config.
- To try other stactools packages, add them to the `worker` requirements. If creating a new custom DAG (like `noaa-hrrr`, either add the package to the `services` requirements, or avoid importing the package in the same file that defines a DAG.


amarouane-ABDELHAK:
@ividito will you be able to demo this for the team?

dags/veda_data_pipeline/veda_stactools_noaa_hrrr.py
```diff
@@ -0,0 +1,99 @@
...
+        stactools_collection = upsert_stactools_collection()
+        ingest_collection = ingest_collection_task(collection=stactools_collection)
+
+        get_items_from_granules = build_items_from_granules()
+        submit_stac = submit_to_stac_ingestor_task_direct.expand(stac_items=get_items_from_granules)
+        submit_stac.set_upstream(ingest_collection)
+
+        get_items_from_granules.set_upstream(start)
+        stactools_collection.set_upstream(start)
+        submit_stac.set_downstream(end)
```
botanical: Out of curiosity, is there any reason to not use the bitwise shift operators? 😯

amarouane-ABDELHAK: +1 for using bitwise shift operators it is more clear

sm2a/airflow_worker/Dockerfile
```diff
@@ -38,10 +38,12 @@ ENV PATH $PATH:/home/airflow/miniconda3/envs/py11/bin
...
 
 COPY --chown=airflow:airflow airflow_worker/requirements.txt "${AIRFLOW_HOME}/requirements.txt"
+COPY --chown=airflow:airflow airflow_worker/requirements-in.txt "${AIRFLOW_HOME}/requirements-in.txt"
 
 RUN pip install --upgrade pip \
-    && pip install "apache-airflow[celery,amazon]==${AIRFLOW_VERSION}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-3.11.txt" \
-    && pip install --no-cache-dir -r requirements.txt -c "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-3.11.txt"
+    && pip install "apache-airflow[celery,amazon]==${AIRFLOW_VERSION}"\
+    && pip install --no-cache-dir -r requirements.txt -c "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-3.11.txt"\
+    && pip install --no-cache-dir -r requirements-in.txt
```
amarouane-ABDELHAK: Why are we installing libraries without constraints?

ividito: I hit a barrier installing `stactools` which requires a more secure version of `httpx`. According to the Airflow [docs](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html), once the Airflow installation is complete, it's recommended that additional dependencies are added as a second step without constraints.

amarouane-ABDELHAK: What do you think about us creating our own constraints.txt? that way we will not have to track two requirements.txt

ividito: I didn't want to spend time untangling it in this PR, but I think if we put the time in, we could end up with something like this:

```
pip install "airflow==${AIRFLOW_VERSION}" -r airflow-providers.txt -c airflow-constraints.txt
pip install dag-requirements.txt -c "airflow==${AIRFLOW_VERSION}"
```

The missing step is splitting up our current requirements file, where it's unclear which dependencies are copy-pasted constraints, and which are needed to maintain DAG functionality. This work can be incremental for a while - new dag dependencies should be added to `requirements.in`, and we can start to prune the old requirements file when we work on #318. Using `uv pip tree` while we do that can help us validate some of the links between dependencies in that file.

botanical (approved):

PR #314, opened 2025-03-12
Fix/dynamic dag args

anayeaye:
## What
Some code updates for clarity (not across the full codebase, just in discovery and vector DAGs)
- Don't ignore a missing bucket arg in the event config. Event.get('bucket') should fail if bucket is not provided so use event['bucket']--this will emit a useful error
- require named parameters

### DIFF
--- a/dags/veda_data_pipeline/groups/discover_group.py
+++ b/dags/veda_data_pipeline/groups/discover_group.py
@@ -14,19 +14,19 @@ group_kwgs = {"group_id": "Discover", "tooltip": "Discover"}
 
 
 @task(retries=1, retry_delay=timedelta(minutes=1))
-def discover_from_s3_task(ti=None, event={}, **kwargs):
+def discover_from_s3_task(*, event: dict = {}, ti=None, payload: dict ={}, prev_start_date_success: str = None):
     """Discover grouped assets/files from S3 in batches of 2800. Produce a list of such files stored on S3 to process.
     This task is used as part of the discover_group subdag and outputs data to EVENT_BUCKET.
     """
-    payload = kwargs.get("payload", ti.dag_run.conf)
+
+    payload = payload or ti.dag_run.conf
     config = {
         **event,
         **payload,
     }
     # TODO test that this context var is available in taskflow
-    last_successful_execution = kwargs.get("prev_start_date_success")
-    if event.get("schedule") and last_successful_execution:
-        config["last_successful_execution"] = last_successful_execution.isoformat()
+    if event.get("schedule") and prev_start_date_success:
+        config["last_successful_execution"] = prev_start_date_success.isoformat()
     # (event, chunk_size=2800, role_arn=None, bucket_output=None):
 
     airflow_vars = Variable.get("aws_dags_variables")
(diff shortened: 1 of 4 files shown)


dags/veda_data_pipeline/utils/s3_discovery.py
```diff
@@ -224,7 +224,7 @@ def s3_discovery_handler(event, chunk_size=2800, role_arn=None, bucket_output=No
         bucket_name=bucket, prefix=prefix, s3_client=s3client
     )
     file_uris = [
-        f"s3://{bucket}/{obj['Key']}"
+        f"s3://{bucket}/{obj['Key']}"
```
anayeaye:
```suggestion
        f"s3://{bucket}/{obj['Key']}"
```

dags/veda_data_pipeline/veda_vector_pipeline.py
```diff
@@ -69,25 +69,23 @@ def ingest_vector_task(payload):
...
-            schedule_interval=params_dag_run_conf.get("schedule", None),
-            params=params_dag_run_conf, 
+            schedule_interval=event.get("schedule", None),
+            params=event, 
             **dag_args
         ) as dag:
         start = DummyOperator(task_id="Start", dag=dag)
         end = DummyOperator(task_id="End", trigger_rule=TriggerRule.ONE_SUCCESS, dag=dag)
-        discover = discover_from_s3_task()
+        discover = discover_from_s3_task(event)
```
anayeaye, line 82: Making the params explicit here made it easier to see why the scheduled DAG was failing without an event but the default user-triggered DAG was working (because the run config is submitted when starting a manual run)

dags/veda_data_pipeline/groups/discover_group.py
```diff
@@ -14,19 +14,19 @@
 
 
 @task(retries=1, retry_delay=timedelta(minutes=1))
-def discover_from_s3_task(ti=None, event={}, **kwargs):
+def discover_from_s3_task(*, event: dict = {}, ti=None, payload: dict ={}, prev_start_date_success: str = None):
```
anayeaye:
```suggestion
def discover_from_s3_task(*, event: dict={}, ti=None, payload: dict={}, prev_start_date_success: str=None):
```

dags/veda_data_pipeline/utils/s3_discovery.py
```diff
@@ -190,7 +190,7 @@ def propagate_forward_datetime_args(event):
 
 
 def s3_discovery_handler(event, chunk_size=2800, role_arn=None, bucket_output=None):
-    bucket = event.get("bucket")
+    bucket = event["bucket"]
     prefix = event.get("prefix", "")
     filename_regex = event.get("filename_regex", None)
```
anayeaye, line 195:
```suggestion
    prefix = event["prefix"]
    filename_regex = event["filename_regex"]
```

dags/veda_data_pipeline/utils/s3_discovery.py
```diff
@@ -190,7 +190,7 @@ def propagate_forward_datetime_args(event):
 
 
 def s3_discovery_handler(event, chunk_size=2800, role_arn=None, bucket_output=None):
-    bucket = event.get("bucket")
+    bucket = event["bucket"]
     prefix = event.get("prefix", "")
```
anayeaye:
```suggestion
    prefix = event["prefix"]
```

smohiudd (approved):

PR #326, opened 2025-03-20
fix(ingest-vector-dag): discover_from_s3 requires event or config object

anayeaye:
## What
The event parameter for the ingest vector dag got dropped in recent changes, this adds it back to support scheduled vector ingest events.

### DIFF
--- a/dags/veda_data_pipeline/veda_vector_pipeline.py
+++ b/dags/veda_data_pipeline/veda_vector_pipeline.py
@@ -115,7 +115,7 @@ def get_ingest_vector_dag(id: str, event: dict):
     ) as dag:
         start = DummyOperator(task_id="Start", dag=dag)
         end = DummyOperator(task_id="End", trigger_rule=TriggerRule.ONE_SUCCESS, dag=dag)
-        discover = start >> discover_from_s3_task(event)
+        discover = start >> discover_from_s3_task(event=event)
         get_files = get_files_task(payload=discover)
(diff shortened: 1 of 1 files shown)


botanical (approved):
I think this works! But I am curious when we choose to use positional arguments vs keyword

anayeaye:
> I think this works! But I am curious when we choose to use positional arguments vs keyword

Good point. I think this is worth event=event for clarity. [This was working previously](https://github.com/NASA-IMPACT/veda-data-airflow/compare/v0.13.5-dev.1...v0.13.5-dev.3#diff-afcc27fa9f6983e0ad8a6de18595805e3c72f77d085f556b1bc9b09b7a44a706L82) but we can improve it.

dags/veda_data_pipeline/veda_vector_pipeline.py
```diff
@@ -115,7 +115,7 @@ def get_ingest_vector_dag(id: str, event: dict):
     ) as dag:
         start = DummyOperator(task_id="Start", dag=dag)
         end = DummyOperator(task_id="End", trigger_rule=TriggerRule.ONE_SUCCESS, dag=dag)
-        discover = start >> discover_from_s3_task()
+        discover = start >> discover_from_s3_task(event)
```
anayeaye:
```suggestion
        discover = start >> discover_from_s3_task(event=event)
```

botanical (approved):

PR #329, opened 2025-03-21
Running Discover and vector DAG without Template

amarouane-ABDELHAK:

We will be able to run discover and vector_ingest DAG with both a scheduled runsand manual runs

### DIFF
--- a/dags/veda_data_pipeline/veda_vector_pipeline.py
+++ b/dags/veda_data_pipeline/veda_vector_pipeline.py
@@ -123,6 +123,6 @@ def get_ingest_vector_dag(id: str, event: dict):
         return dag
 
 
-# Sending empty event because we rely on task instance (ti) for manula runs
+# Sending empty event because we rely on task instance (ti) for manual runs
 # and payload for scheduled runs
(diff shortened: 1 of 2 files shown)


dags/veda_data_pipeline/veda_vector_pipeline.py
```diff
@@ -117,9 +118,11 @@ def get_ingest_vector_dag(id: str, event: dict):
...
         discover = start >> discover_from_s3_task(event=event)
         get_files = get_files_task(payload=discover)
-        ingest_vector_task.expand(payload=get_files) >> invalidate_cloudfront(event=event) >> end
+        ingest_vector_task.expand(payload=get_files) >> invalidate_cloudfront() >> end
 
         return dag
 
 
-get_ingest_vector_dag(id="veda_ingest_vector", event=template_dag_run_conf)
+# Sending empty event because we rely on task instance (ti) for manula runs
```
anayeaye:
```suggestion
# Sending empty event because we rely on task instance (ti) for manual runs
```

anayeaye (approved):

PR #302, opened 2025-02-27
Add RDS disaster recovery

paridhi-parajuli:
This PR adds two DAGs, get_snapshots_dag and rds_s3_export, which manage the retrieval and export of RDS snapshots to S3, along with metadata indexing using AWS Glue.

Related tickets:
https://github.com/US-GHG-Center/ghgc-architecture/issues/429
https://github.com/US-GHG-Center/ghgc-architecture/issues/495

**Tests:**
Tested in `ghgc-smce-dev` for `rds veda-backend-ghgc-dev-postgres`, creates snapshots in `ghg-rds-snapshots-dev` and the corresponding database in AWS Glue.

Queries through Athena:
SELECT * FROM "veda-backend-ghgc-dev-postgres".postgis LIMIT 10;
Works fine!

PR #332, opened 2025-03-21
Reopen "Add RDS disaster recovery"

ividito:
Brings back #302


sm2a/infrastructure/variables.tf
```diff
@@ -188,9 +215,24 @@ variable "assume_role_write_arn" {
   default = ""
 }
 
+
+
+variable "snapshot_bucket_name" {
+  default = "ghg-rds-snapshots-dev"
+}
```
ividito: Can we default this to empty?

sm2a/infrastructure/variables.tf
```diff
@@ -98,6 +98,33 @@ variable "custom_worker_policy_statement" {
...
+      "Effect" : "Allow",
+      "Action" : [
+        "glue:DeleteDatabase"
+      ],
+      "Resource" : [
+        "arn:aws:glue:us-west-2:*:catalog",
+        "arn:aws:glue:us-west-2:*:database/*",
+        "arn:aws:glue:us-west-2:*:table/*",
+        "arn:aws:glue:us-west-2:*:userDefinedFunction/*"
+      ]
```
ividito, line 127: policy is invalid

sm2a/infrastructure/rds_dr_roles.tf
```diff
@@ -0,0 +1,130 @@
...
+        Action = [
+          "s3:PutObject*",
+          "s3:ListBucket",
+          "s3:GetObject*",
+          "s3:DeleteObject*",
+          "s3:GetBucketLocation"
+        ]
+        Resource = [
+          "arn:aws:s3:::${var.snapshot_bucket_name}",
+          "arn:aws:s3:::${var.snapshot_bucket_name}/*"
```
ividito, line 57: These might need to be modified if this is an optional or non-defaulted value

ividito (commented):
A few things in this PR should be changed before merging. Some are minor TF syntax changes to maintain valid deployments. I think the more complicated part is maintaining support for non-GHG instances by making some of the new resources and variables optional.

sm2a/infrastructure/rds_dr_roles.tf
paridhi-parajuli: The snapshot bucket name is always required

paridhi-parajuli (approved):

PR #336, opened 2025-03-24
fix(dag-generation): allow empty optional datetime_range 

anayeaye:
## What
Resolve param validation error when config does not include the optional datetime_range parameter


anayeaye:
@ividito @amarouane-ABDELHAK I am not sure the best way to handle optional airflow Params. Datetime_range is not required for either COG or vector discovery.

ividito (approved):

PR #337, opened 2025-03-24
fix(generate-dags): allow empty string in optional datetime range enums

anayeaye:
## What

Try to support an optional airflow Param with ENUMs list. `None` is not type string so an empty string may get through airflow and hopefully will be properly handled in the DAG runtime


anayeaye:
@ividito @amarouane-ABDELHAK The pythonic way (None) did not work. Perhaps using an empty string will work for an optional parameter with an enum?

ividito (approved):

Issue #300, opened 2025-02-19
Transfer task should not fail when files have been previously transferred


anayeaye:
## What
The transfer task attempts to not duplicate transfers by checking if the file has previously been transferred ([veda_data_pipeline/utilstransfer.py](https://github.com/NASA-IMPACT/veda-data-airflow/blob/dev/dags/veda_data_pipeline/utils/transfer.py#L44)). Unfortunately this causes the pipeline to fail--the pipeline should go on to generate and publish STAC metadata regardless of whether or not the files have previously been transferred.

## Note
When transfer is set to false in a promotion config, the pipeline fails
```
ip-10-41-200-63.us-west-2.compute.internal
*** Found logs in s3:
***   * s3://veda-tf-state-production/remote_base_log_folder/dag_id=veda_promotion_pipeline/run_id=manual__2025-02-19T21:50:49+00:00/task_id=transfer_assets_to_production_bucket/map_index=0/attempt=1.log
[2025-02-19, 21:52:36 UTC] {taskinstance.py:1979} INFO - Dependencies all met for dep_context=non-requeueable deps ti=<TaskInstance: veda_promotion_pipeline.transfer_assets_to_production_bucket manual__2025-02-19T21:50:49+00:00 map_index=0 [queued]>
[2025-02-19, 21:52:36 UTC] {taskinstance.py:1979} INFO - Dependencies all met for dep_context=requeueable deps ti=<TaskInstance: veda_promotion_pipeline.transfer_assets_to_production_bucket manual__2025-02-19T21:50:49+00:00 map_index=0 [queued]>
[2025-02-19, 21:52:36 UTC] {taskinstance.py:2193} INFO - Starting attempt 1 of 2
[2025-02-19, 21:52:36 UTC] {taskinstance.py:2217} INFO - Executing <Mapped(_PythonDecoratedOperator): transfer_assets_to_production_bucket> on 2025-02-19 21:50:49+00:00
[2025-02-19, 21:52:36 UTC] {standard_task_runner.py:60} INFO - Started process 2390 to run task
[...]

PR #342, opened 2025-04-02
fix: when transfer is false in a promotion config, do not transfer files. when files are up to date, continue 

botanical:

Addresses https://github.com/NASA-IMPACT/veda-data-airflow/issues/300

## Changes

* updated `transfer_files_within_s3` to check transfer setting before attempting to transfer files
* set transfer default to true
* when files are up to date and (PreconditionFailed) error is received, continue without erroring. 

## PR Checklist

- [ ] Update CHANGELOG
- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/dags/veda_data_pipeline/utils/transfer.py
+++ b/dags/veda_data_pipeline/utils/transfer.py
@@ -42,12 +42,15 @@ def get_matching_files(s3_client, bucket, prefix, regex_pattern):
 
 
 def transfer_files_within_s3(
-    s3_client, origin_bucket, matching_files, destination_bucket, collection
+    s3_client, origin_bucket, matching_files, destination_bucket, collection, transfer=True
 ):
+    if not transfer:
+        print(f"Transfer is disabled. Would have copied {len(matching_files)} files from {origin_bucket} to {destination_bucket}")
+        return
+
     transfer_exceptions = False
     for file_key in matching_files:
         filename = file_key.split("/")[-1]
-        # print(f"Transferring file: {filename}")
         target_key = f"{collection}/{filename}"
         copy_source = {"Bucket": origin_bucket, "Key": file_key}
 
@@ -91,6 +94,7 @@ def data_transfer_handler(event, role_arn=None):
     filename_regex = event.get("filename_regex")
     target_bucket = event.get("target_bucket")
     collection = event.get("collection")
+    transfer = event.get("transfer")
 
     kwargs = assume_role(role_arn=role_arn) if role_arn else {}
     s3client = boto3.client("s3", **kwargs)
@@ -111,6 +115,7 @@ def data_transfer_handler(event, role_arn=None):
             matching_files=matching_files,
             destination_bucket=target_bucket,
             collection=collection,
+            transfer=transfer
         )
     else:
         print(
--- a/dags/veda_data_pipeline/utils/transfer.py
+++ b/dags/veda_data_pipeline/utils/transfer.py
@@ -56,6 +56,7 @@ def transfer_files_within_s3(
 
         # We can use the etag to check if the file has already been copied and avoid duplication of effort
         # by using the CopySourceIfNoneMatch parameter below.
+        target_etag = None
         try:
             target_metadata = s3_client.head_object(
                 Bucket=destination_bucket, Key=target_key
@@ -76,6 +77,11 @@ def transfer_files_within_s3(
                     Bucket=destination_bucket,
                     Key=target_key
                 )
+            elif err.response["Error"]["Code"] == "PreconditionFailed":  # 412 error
+                # File is already up to date, skip copying
+                print(f"File {filename} is already up to date, skipping")
+                print(f"CopySourceIfNoneMatch: {target_etag}. Skip copy: object is unchanged (ETag matches).")
+                continue
             else:
                 msg = f"ClientError copying {filename=} {err=}"
                 print(msg)
--- a/dags/veda_data_pipeline/utils/transfer.py
+++ b/dags/veda_data_pipeline/utils/transfer.py
@@ -42,12 +42,8 @@ def get_matching_files(s3_client, bucket, prefix, regex_pattern):
 
 
 def transfer_files_within_s3(
-    s3_client, origin_bucket, matching_files, destination_bucket, collection, transfer
+    s3_client, origin_bucket, matching_files, destination_bucket, collection
 ):
-    if not transfer:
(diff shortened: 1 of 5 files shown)


dags/veda_data_pipeline/utils/transfer.py
```diff
@@ -42,12 +42,15 @@ def get_matching_files(s3_client, bucket, prefix, regex_pattern):
 
 
 def transfer_files_within_s3(
-    s3_client, origin_bucket, matching_files, destination_bucket, collection
+    s3_client, origin_bucket, matching_files, destination_bucket, collection, transfer
```
ividito: In the DAG docs, we say this parameter defaults to false. Should we enforce that behaviour in this method? (or in the `transfer_assets_to_production_bucket` [task](https://github.com/NASA-IMPACT/veda-data-airflow/blob/e2d74c472868666e005e7cae2c8e9024506b9d2a/dags/veda_data_pipeline/veda_promotion_pipeline.py#L83) )

ividito (commented):
Minor comment. I'm also interested to know whether you tried [filtering out ](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services)the `PreconditionFailed` errors - I think it's a good idea to preempt the transfer attempts, but it might complicate our s3 data management later on.

```
except botocore.exceptions.ClientError as err:
    if error.response['Error']['Code'] == 'PreconditionFailed':
        ...
    else:
        ...
```

botanical:
> Minor comment. I'm also interested to know whether you tried [filtering out ](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html#parsing-error-responses-and-catching-exceptions-from-aws-services)the `PreconditionFailed` errors - I think it's a good idea to preempt the transfer attempts, but it might complicate our s3 data management later on.
> 
> ```
> except botocore.exceptions.ClientError as err:
>     if error.response['Error']['Code'] == 'PreconditionFailed':
>         ...
>     else:
>         ...
> ```

@ividito I didn't try to filter out the `PreconditionFailed` errors. Do you think it might be worthwhile for us to consider refreshing the object metadata and retrying a copy if we get a 412? Why do we have transfer as a parameter of the DAG if we want to try to copy/transfer objects regardless of the setting? 😯

ividito:
>Do you think it might be worthwhile for us to consider refreshing the object metadata and retrying a copy if we get a 412? 

I think this only solves the issue of trying to transfer a file that has already been copied over - our only precondition is a matching `etag`, so if that error comes through in the exception, we can safely ignore it.

dags/veda_data_pipeline/veda_promotion_pipeline.py
```diff
@@ -92,8 +96,8 @@ def transfer_assets_to_production_bucket(ti=None, payload={}):
     end = EmptyOperator(task_id="end", dag=dag)
 
     collection_grp = collection_task_group()
-    mutate_payload_task = remove_thumbnail_asset()
-    extract_from_payload = extract_discovery_items_from_payload()
+    mutate_payload_task = remove_thumbnail_asset(ti=None)
+    extract_from_payload = extract_discovery_items_from_payload(ti=None)
```
ividito: I don't think this is needed, `ti` is an [automatically-populated ](https://airflow.apache.org/docs/apache-airflow/2.10.5/tutorial/taskflow.html#accessing-context-variables-in-decorated-tasks)`kwarg`/context variable that comes from the task decorator
```suggestion
    mutate_payload_task = remove_thumbnail_asset()
    extract_from_payload = extract_discovery_items_from_payload()
```

ividito (changes requested):
One comment, but this is great otherwise!

dags/veda_data_pipeline/veda_promotion_pipeline.py
anayeaye: I had some trouble snipping `ti=None` out of some sub tasks. I agree it _shouldn't_ be needed but I have been unable to get around it in some cases. Probably indicates something about flow but sorting that out may be a bigger project

botanical: I figured it out! I didn't have the right version of airflow installed. After upgrading to the version we're using (2.8.4) and also installing the correct version of pluggy>=1.5.0, I was able to get the unit tests working without `ti=None` set.

dags/veda_data_pipeline/veda_promotion_pipeline.py
```diff
@@ -18,30 +20,31 @@
...
+    "title": "collection-title",
+    "transfer": "false"
 }
 ```
 """
 
 dag_args = {
     "start_date": pendulum.today("UTC").add(days=-1),
-    "schedule_interval": None,
+    "schedule": None,
```
botanical, line 47: @ividito Is it okay to update this? `schedule_interval` is deprecated. https://github.com/apache/airflow/pull/25410

https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#deprecation-of-schedule-interval-and-timetable-arguments-25410

dags/veda_data_pipeline/veda_promotion_pipeline.py
```diff
@@ -1,8 +1,10 @@
 import pendulum
 from airflow import DAG
 from airflow.decorators import task
-from airflow.operators.dummy_operator import DummyOperator as EmptyOperator
+from airflow.operators.empty import EmptyOperator
```
anayeaye, line 4: Nice. I was wondering about this, we can follow suit when updating other pipelines
> [airflow.operators.dummy](https://airflow.apache.org/docs/apache-airflow/2.3.2/_api/airflow/operators/dummy/index.html#module-airflow.operators.dummy)
This module is deprecated. Please use [airflow.operators.empty](https://airflow.apache.org/docs/apache-airflow/2.3.2/_api/airflow/operators/empty/index.html#module-airflow.operators.empty).

dags/veda_data_pipeline/utils/transfer.py
anayeaye: After reading the flow a few times, I agree that this might be smoother to enforce in the transfer_assets_to_production task. Also we sometimes manually trigger the transfer DAG from the UI and it feels odd to add transfer T/F to the run configuration for the transfer DAG.

botanical: Updated https://github.com/NASA-IMPACT/veda-data-airflow/pull/342/commits/c5de93ba7ebe69956001c95aaf6cbdaea1fa4e93 !

ividito (approved):

anayeaye (approved):

Issue #325, opened 2025-03-20
Improve failure handling for bulk ingestion to prevent item failures from blocking valid items


abarciauskas-bgse:
Description:
* Currently, the dataset pipeline fails to complete ingestion if any single file encounters an issue, even when the majority of files are valid. 

Context:
* Recent OCO-3 workflow runs highlighted this issue when only 2 out of 98 files failed, but the entire ingestion process was halted due to "upstream failed" status in Airflow.
* Similar failures have been encountered in other bulk ingests, particularly when files contain bands with 100% no data.

Proposed Actions:
* Investigate if Airflow DAGs can be configured to continue processing even if some items fail.
* Explore adding an option to retry or skip failed items instead of blocking the entire ingestion.
* Consider logging and flagging failed items separately instead of marking the entire batch as unsuccessful.

cc @ividito @anayeaye @smohiudd

anayeaye:
I backed out my earlier work. The new plan should be to remove the raised exception in the build stac step and add reporting. See CSDA reporting example https://github.com/NASA-IMPACT/csda-self-managed-airflow/blob/571c70bc703850674b4e40aa64837b1a268a13cf/dags/restore_dag.py#L264 also see automation DAG reporting in veda-data-airflow.

smohiudd:
Prior work:
https://github.com/NASA-IMPACT/veda-data-airflow/pull/333
https://github.com/NASA-IMPACT/veda-data-airflow/pull/327

anayeaye:
https://github.com/NASA-IMPACT/veda-data-airflow/pull/380

PR #343, opened 2025-04-07
fix(dataset dag): run submit stack when all build stack tasks are done not all_success

anayeaye:
## What
Add `all_done` [trigger rule](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html#trigger-rules) to submit stac step of dataset ingestion pipeline to unblock publishing metadata for valid S3 objects when STAC metadata can be generated at least one discovered S3 object.

1. Run submit stac task when all build stack tasks are done regardless of success
    - The [payload emitted by the build stac task ](https://github.com/NASA-IMPACT/veda-data-airflow/blob/dev/dags/veda_data_pipeline/utils/build_stac/handler.py#L108-L133)separates successful metadata generation from failed metadata generation
    - Submit stac only processes the [successfully generated item metadata from the build stac task](https://github.com/NASA-IMPACT/veda-data-airflow/blob/dev/dags/veda_data_pipeline/groups/processing_tasks.py#L38-L54)

2. Only raise `ValueError` when no metadata was successfully generated (instead of if any item failed in the batch)

## Why
https://github.com/NASA-IMPACT/veda-data-airflow/issues/325

## How to test
When deployed to dev we can try rerunning the dataset ingestion in #325 that is currently failing before any metadata can be published if a single un-publishable item is identified in the discovery step. Our goal is to report what S3 objects caused failures while still publishing metadata for all the other objects discovered.

## NOTE
This work only partially resolves the #325 issue. After confirming that this workflow is unblocked, we will still need to add error reporting and slack notifications

### DIFF
--- a/dags/veda_data_pipeline/veda_dataset_pipeline.py
+++ b/dags/veda_data_pipeline/veda_dataset_pipeline.py
@@ -47,11 +47,11 @@ dag_args = {
 
 with DAG("veda_dataset_pipeline", params=template_dag_run_conf, **dag_args) as dag:
     start = EmptyOperator(task_id="start")
-    end = EmptyOperator(task_id="end")
+    end = EmptyOperator(task_id="end", trigger_rule=TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS)
 
     mutated_payloads = start >> collection_task_group() >> remove_thumbnail_asset()
     discovery_items = extract_discovery_items_from_payload(payload=mutated_payloads)
     discover = discover_from_s3_task.partial(payload=mutated_payloads).expand(event=discovery_items)
     get_files = get_files_task(payload=discover)
     build_stac = build_stac_task.expand(payload=get_files)
-    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac, trigger_rule=TriggerRule.ALL_DONE) >> end
+    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac) >> end
--- a/dags/veda_data_pipeline/veda_dataset_pipeline.py
+++ b/dags/veda_data_pipeline/veda_dataset_pipeline.py
@@ -1,6 +1,7 @@
 import pendulum
 from airflow import DAG
 from airflow.models.param import Param
+from airflow.utils.trigger_rule import TriggerRule
 from veda_data_pipeline.groups.discover_group import discover_from_s3_task, get_files_task
 from airflow.operators.dummy_operator import DummyOperator as EmptyOperator
 from veda_data_pipeline.groups.collection_group import collection_task_group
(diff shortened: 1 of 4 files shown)


dags/veda_data_pipeline/veda_dataset_pipeline.py
```diff
@@ -53,4 +54,4 @@
     discover = discover_from_s3_task.partial(payload=mutated_payloads).expand(event=discovery_items)
     get_files = get_files_task(payload=discover)
     build_stac = build_stac_task.expand(payload=get_files)
-    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac) >> end
+    submit_stac = submit_to_stac_ingestor_task.expand(built_stac=build_stac, trigger_rule=TriggerRule.ALL_DONE) >> end
```
ividito:
```suggestion
    submit_stac = submit_to_stac_ingestor_task.partial(trigger_rule=TriggerRule.ALL_DONE).expand(built_stac=build_stac) >> end
```

anayeaye: @ividito are you suggesting partial because the trigger rule may not be applied to all tasks in expand? I don't quite see why we need it here

https://airflow.apache.org/docs/apache-airflow/stable/authoring-and-scheduling/dynamic-task-mapping.html#adding-parameters-that-do-not-expand

ividito: I'm not certain what the behaviour is with the trigger rule expanded. The way I read it is that using `expand`, it gets unpacked as an iterable, with each value used to map a new task. If the trigger rule is set using `partial`, then it doesn't get expanded - it's used as a constant parameter for all the tasks created using `expand`. 

We have a (not great) example on the `discover` task, where we map over the discovery objects, but use `partial` to keep a constant set of collection-level information across all mapped tasks.

anayeaye: Looking at adding this to the DummyOperator and what that might mean...

`TypeError: partial() got an unexpected keyword argument 'trigger_rule'` <- same import error happens when adding trigger rule to expand

anayeaye: Screenshot 2025-04-08 at 3 52 23 PM

This was a local run using `end = EmptyOperator(task_id="end", trigger_rule=TriggerRule.ALL_DONE)`. The early collection task failure is expected but I think for this case it should be a fail so I'm looking at the other trigger rule options so we won't always report successful execution 🤔

anayeaye: @ividito I moved the trigger rule to the submit stack task--what do you think? I still feel like this is an odd way to control the flow so I'll revisit your pr #333

anayeaye:
Closing after reviewing with the team. Instead of overriding default trigger behavior we will instead stop raising errors on individual build stac failures and at the same time we will add reporting to surface failures

abarciauskas-bgse:
Thanks for working on this @anayeaye and @ividito !

Issue #348, opened 2025-04-11
Remove MWAA terraform infrastructure


amarouane-ABDELHAK:
**Description**:
As we plan to retire MWAA, we need to remove all associated Infrastructure as Code (IaC) from the repository.

**Acceptance Criteria**:
- [ ] All MWAA-related infrastructure is fully removed from the veda-data-airflow repository
- [ ] GitHub Actions continue to deploy SM2A as before, without any issues

PR #351, opened 2025-04-15
feat: Start using uv

ividito:
**Summary:**

Reduces dev dependency installation time by 99% (47.8s on my machine to 400ms - at first, I didn't believe `uv sync` had actually done anything at all).

I tried to more-or-less follow conventions established by CSDA: https://github.com/NASA-IMPACT/csdap-stac-api/pull/250

Included some minor README updates to match.

Starts some of the dependency/constraint reorganization discussed in #307 - I would like to get this working in the Dockerfiles too, but this will be a lot easier after #348


README.md
```diff
@@ -35,6 +27,10 @@ See [terraform-getting-started](https://developer.hashicorp.com/terraform/tutori
 
 See [getting-started-install](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html)
 
+### Python
+
+- Install [uv](https://docs.astral.sh/uv/getting-started/installation/).
+- Run `uv sync` to install the required python packages. By default, all optional dependencies are included. To avoid this, use `uv sync --no-default-groups`.
```
botanical, line 33: What version of uv do we need? 

I'm seeing:

```
uv sync --no-default-groups                                                                     
error: unexpected argument '--no-default-groups' found

  tip: a similar argument exists: '--no-all-extras'

Usage: uv sync [OPTIONS]

For more information, try '--help'.
```

on `uv 0.3.2 (c5440001c 2024-08-23)`. 

And running `uv sync` gives me:
```
warning: Failed to parse `pyproject.toml` during settings discovery:
  TOML parse error at line 87, column 1
     |
  87 | [tool.uv]
     | ^^^^^^^^^
  unknown field `default-groups`

Failed to parse lockfile; ignoring locked requirements: TOML parse error at line 3810, column 1
     |
3810 | [[package]]
     | ^^^^^^^^^^^
data did not match any variant of untagged enum SourceWire
Resolved 134 packages in 2.87s
error: Failed to prepare distributions
  Caused by: Failed to fetch wheel: veda-data-airflow @ file:///Users/jennifertran/Code/ds/VEDA/veda-data-airflow
  Caused by: Build backend failed to determine extra requires with `build_editable()` with exit status: 1
--- stdout:

--- stderr:
error: Multiple top-level packages discovered in a flat-layout: ['etc', 'img', 'mnt', 'opt', 'dev', 'proc', 'home', 'dags', 'boot', 'sm2a', 'root', 'media', 'routes', 'docker_tasks', 'workflows_api', 'infrastructure', 'data_airflow_env'].

To avoid accidental inclusion of unwanted files or directories,
setuptools will not proceed with this build.

If you are trying to create a single distribution with multiple packages
on purpose, you should not rely on automatic discovery.
Instead, consider the following options:

1. set up custom discovery (`find` directive with `include` or `exclude`)
2. use a `src-layout`
3. explicitly set `py_modules` or `packages` with a list of names

To find more information, look for "package discovery" on setuptools docs.
```

botanical (commented):
Yay for uv!! I just had one question.

README.md
ividito: Dependency groups were [added with 0.4.27](https://github.com/astral-sh/uv/blob/main/changelogs/0.4.x.md#0427), using that version also fixes the other error you had. 

`uv self update` will bring you to the latest version, and it's reversible with `uv self update <version number>`

amarouane-ABDELHAK:
I am a huge fan of `uv` and I am in favor of using it. But, SM2A is intended to run in a Docker container for local development to mirror the production environment (as closely as possible). Is the use case for uv here specifically to run unit tests?
After running uv sync, there’s no further guidance on what the user should do next with the virtual environment set up by uv.

ividito:
@amarouane-ABDELHAK for now, yes, this only improves non-containerized local development and unit testing. I have a follow-up that incorporates `uv` into our Dockerfiles, which will have more impact on deployment speed and containerized development, but I think this will be easier to implement and merge after #352

botanical (changes requested):
Since we're using an updated version of moto, do you mind updating the unit tests that import it? ( I think the `mock_s3` imports can now be `from moto import mock_aws` and all instances of `mock_s3` can be updated to use `mock_aws`.

I also think we need to add `smart_open` to our `pyproject.toml`

botanical (approved):
🔥 thanks for this work

PR #357, opened 2025-04-17
Revise constraints to keep xmlsec from updating

ividito:
Small fix - just making sure Airflow dependencies are pinned properly (this began to raise errors after `xmlsec` pushed a broken update)


amarouane-ABDELHAK (approved):
Looks good, but with some comments:
- We should note that we might face a dependencies conflict since we are installing some libraries without constraints 
- We really don't need to install `stac_tools` in the services containers since it will be used only for the DAGs

PR #361, opened 2025-04-25
Add support for pyarc2stac based ingest

kyle-lesinger:
## **Summary:** 

1.) Create a DAG that create a collection using pyarc2stac
2.) Make this DAG schedulable
2.) Install pyarc2stac within airflow_worker and airflow_services within virtual environment (because of conflict with some library on noaa-hrr stac tools dag (we should do this for all the dags - libraries specifically only needed by the dag needs to go within the dag specific virtual environment)

## PR Checklist

- [ ] Integration tests


slesaad:
@kyle-lesinger thanks for starting the work on this.

Ideally, we want to have a generic DAG (that has nothing to do with disasters) that's not always running on schedule

Scenario: I want to ingest an ArcGIS dataset for the air quality (or some other) portal and run it only once. The ideal solution should also cover this scenario. It's kinda getting there with you reading the json from an s3 bucket, but not completely.

If you don't know this already, our airflow instance comes with an s3 bucket that saves the dags in the `/dags` folder. The bucket is available through the following env var: `MWAA_STAC_CONF["EVENT_BUCKET"]`. The same s3 bucket has been used to hold scheduled datasets in the `/collections` folder. Airflow knows to look into that folder and read the json from there, and based on the schedule provided, run the discover dag for the dataset in schedule. This feature was added in https://github.com/NASA-IMPACT/veda-data-airflow/pull/155, you can check it out there.

Here's what I'd do:
1. remove all reference of "disasters" - we want a generic solution that works for everything
2. update the dag so that any config can be run manually
3. handled scheduled workflow for pyarc2stac by updating the `generate_dags` function

i'd accept the input config that looks something like this:

{
  "url": "https://arcgis.url/no2/imageserver",
  "id": "no2",
  "description": "description of no2 data",
  ....(any other stac collection compatible properties that you'd like to overwrite in the collection created by pyarc2stac)
}

let me know if you wanna meet in person to go over the comments.

slesaad (changes requested):
@kyle-lesinger thanks for starting the work on this.

Ideally, we want to have a generic DAG (that has nothing to do with disasters) that's not always running on schedule

Scenario: I want to ingest an ArcGIS dataset for the air quality (or some other) portal and run it only once. The ideal solution should also cover this scenario. It's kinda getting there with you reading the json from an s3 bucket, but not completely.

If you don't know this already, our airflow instance comes with an s3 bucket that saves the dags in the `/dags` folder. The bucket is available through the following env var: `MWAA_STAC_CONF["EVENT_BUCKET"]`. The same s3 bucket has been used to hold scheduled datasets in the `/collections` folder. Airflow knows to look into that folder and read the json from there, and based on the schedule provided, run the discover dag for the dataset in schedule. This feature was added in https://github.com/NASA-IMPACT/veda-data-airflow/pull/155, you can check it out there.

Here's what I'd do:
1. remove all reference of "disasters" - we want a generic solution that works for everything
2. update the dag so that any config can be run manually
3. handled scheduled workflow for pyarc2stac by updating the `generate_dags` function

i'd accept the input config that looks something like this:
```
{
  "url": "https://arcgis.url/no2/imageserver",
  "id": "no2",
  "description": "description of no2 data",
  ....(any other stac collection compatible properties that you'd like to overwrite in the collection created by pyarc2stac)
}
```
let me know if you wanna meet in person to go over the comments.

dags/generate_dags.py
```diff
@@ -89,4 +90,13 @@ def generate_dags():
...
             )
 
+        # veda_pyarc2stac_ingest
+        scheduled_pyarcstac_configs = filter_configs_by_dag(collection_configs, "veda_pyarc2stac_ingest")
+        
+        for idx, vector_config in enumerate(scheduled_pyarcstac_configs):
+            id = f"pyarc2stac-{vector_config['id']}"
+            get_ingest_pyarc2stac_dag(
+                id=id, event=vector_config
+            )
```
slesaad:
```suggestion
        for idx, pyarc2stac_config in enumerate(scheduled_pyarcstac_configs):
            id = f"pyarc2stac-{pyarc2stac_config['id']}"
            get_ingest_pyarc2stac_dag(
                id=id, event=pyarc2stac_config
            )
```

dags/generate_dags.py
```diff
@@ -89,4 +90,13 @@ def generate_dags():
                 id=id, event=vector_config
             )
 
+        # veda_pyarc2stac_ingest
+        scheduled_pyarcstac_configs = filter_configs_by_dag(collection_configs, "veda_pyarc2stac_ingest")
```
slesaad: where is this name `veda_pyarc2stac_ingest` defined? 🤔

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
```diff
@@ -0,0 +1,150 @@
...
+    for key in keys_to_overwrite:
+        event_value = event.get(key)
+        if isinstance(event_value, str) and event_value.strip():
+            collection[key] = event_value
+
+
+    # Overwrite collection values with template configuration values. This will only overwrite if the value is a non-empty string within the template_conf file.
+    for key, value in template_conf.items():
+        if isinstance(value, str) and value.strip():
+            collection[key] = value
```
slesaad: i'm unsure why two of these are needed

dags/generate_dags.py
kyle-lesinger: It's located within the veda-tf-state-shared/collections/veda-pyarc2stac.json as a key/value pair

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
kyle-lesinger: To increase the flexibility of the code, these cover the options to a.) use only information from the AWS .json file, b.) use information from the manual triggering. Pyarc2stac generates its own information but I foresee that having only the .json file content overwriting the pyarc2stac outputs cannot account for manual triggering and changes in the configuration file. 

Just for clarity: The "event" is from the .json file, while the "template_conf" is from SM2A airflow.

dags/generate_dags.py
kyle-lesinger: I just noticed the commit suggestion button. I have pushed these requested changes.

dags/generate_dags.py
slesaad: Ah, I thought the mapping to a dag was automatic based on the dag ID rather than manually. Maybe we can update this, though. Instead of having three different for loops, have one and just map the configs to the respective dag based on the dag ID.

I was wondering if you're missing something like this for the pyarc2stac dag https://github.com/NASA-IMPACT/veda-data-airflow/blob/8ff9fc9eb4639aceedf37e374e62d154e2329ad1/dags/veda_data_pipeline/veda_discover_pipeline.py#L101

kyle-lesinger: @slesaad I've now creating a mapping function for each pipeline type. Additionally, I've added `get_ingest_pyarc2stac_dag(id="veda_pyarc2stac_ingest", event={})` per recommendation to account for manual runs.

dags/generate_dags.py
```diff
@@ -8,25 +8,46 @@
...
+    for idx, collection in enumerate(collection_configs):
+        if not collection.get("schedule"):
+            continue
+
+        dag_key = collection.get("dag", "veda_discover")
+        builder, prefix = dag_configs[dag_key]
+
+        # Rename the task_id if the collection has an "id" field
+        if prefix == "pyarc2stac":
+            id = f"{prefix}-{collection['id']}"
```
slesaad: is this really necessary? why not follow the pattern that's already there? (then you wont need the dag_config values to be a tuple) and if you feel strongly, update all of them to use collection id?

dags/generate_dags.py
```diff
@@ -39,6 +60,19 @@ def generate_dags():
...
     bucket = airflow_vars_json.get("EVENT_BUCKET")
 
+    '''Define the mapping of DAG builders to their respective keys and prefixes
+    The key values (e.g., veda_discover) are located as a key value pair in the AWS S3 bucket under the collections/ folder in the .json file.
+    The mapping functions are located in the veda_data_pipeline directory (tuple index 0).
+    The naming ID (e.g., discover, vector, pyarc2stac) is used to generate an id name for each DAG (tuple index 1).
+    '''
+
+    dag_configs = {
+        "veda_discover":          (get_discover_dag,      "discover"),
```
slesaad: you don't really need the values to be a tuple. `my_function.__name__
` should give you the name of the function as a string and you should be able to do any conditional based on that

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
slesaad: not sure if that's needed. i think the `veda_discover_pipeline` does the same thing without needing to account for the two different options, can you confirm if this is absolutely needed?

dags/generate_dags.py
kyle-lesinger: It's only necessary if we want the names of the collection being ingested to be understandable within airflow. If we only set it to ` id = f"{prefix}-{file_name}"` then we will have the dags created as pyarc2stac-veda-pyarcstac-1, pyarc2stac-veda-pyarcstac-2, pyarc2stac-veda-pyarcstac-3, etc.

It only assists to reduce the overhead during debug if one of the ingests fails. But it's not needed at all.  So I'll let you decide what works best. Just let me know.

dags/generate_dags.py
kyle-lesinger: Yes that makes sense. I was trying to allow for different naming conventions to be created, but I'll simply draw the name from the string.

dags/generate_dags.py
kyle-lesinger: image

slesaad: that makes sense, let's use this pattern for the others too?

dags/generate_dags.py
slesaad: you can also use the dictionary key itself (`dag_key`) - has the same information

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
```diff
@@ -0,0 +1,154 @@
...
+- This DAG can run with the following configuration <br>
+```json
+{
+    "url": "https://maps.disasters.nasa.gov/ags03/rest/services/NRT/lis_ak_green_veg_fraction/ImageServer",
+    "stac_id": "nrt_lis_ak_green_veg_fraction",
+    "title": "collection-title",
+    "stac_version": "1.0.0",
+    "description": "collection description",
+    "data_type": "",
+    "is_periodic": true,
```
anayeaye: Note: We _may_ need to add a prefix here and in the template. The dashboard prefixed properties `dashboard:is_periodic` and `dashboard:time_density` have been handled different ways--in our deprecated workflows API `is_periodic` is an alias for `dashboard:is_periodic` but I don't think we support that alias right now

I will raise this in tag up and follow up here--I think the we may need to make adjustments to all of our DAGs that generate collection metadata (not just this new DAG).
cc: @slesaad

dags/generate_dags.py
kyle-lesinger: Complete. Final function call is:
```    dag_configs = {
        "veda_discover":          get_discover_dag,
        "veda_ingest_vector":     get_ingest_vector_dag,
        "veda_pyarc2stac_ingest": get_ingest_pyarc2stac_dag,
    }
```

And the name comes directly from the function `__name__`.

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
kyle-lesinger: Added dashboard to the config file:
```
template_conf = {
    "url": "",
    "id": "",
    "title": "",
    "stac_version": "",
    "description": "",
    "data_type": "",
    "license": "",
    "dashboard:is_periodic": "",
    "dashboard:time_density": ""
}
```

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
kyle-lesinger: I've added new functionality to allow for a specific order of collection values. 
- Manual config triggering = precedence 1 
- AWS .json config. = precedence 2
- pyarc2stac outputs = precedence 3 

```
    for key in collection.keys():
        collection[key] = (template_conf.get(key) or event.get(key) or collection[key])
```

dags/generate_dags.py
kyle-lesinger: After further updates, this loop has been abandoned.

slesaad:
@kyle-lesinger looks pretty good to be, let's resolve the conflicts and then we'll merge

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
```diff
@@ -0,0 +1,193 @@
...
+        raise ValueError(
+            "URL is required but not provided in the event or template_conf."
+        )
+
+    # Generate STAC collection from ArcGIS Image/Map/Feature Server
+    reader = ArcReader(server_url=url)
+    collection = reader.generate_stac().to_dict()
+
+
+    def _choose_keyValues(key, default):
```
kyle-lesinger, line 103: This function assists with actually overwriting values based on either the manual ingest, the .json payload in AWS, or the pyarc2stac output.

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
```diff
@@ -0,0 +1,193 @@
...
+        and `dashboard:is_periodic`.
+        """
+        value = template_conf.get(key)
+        if value not in (None, ""):
+            return value
+
+        value = event.get(key)
+        return value if value not in (None, "") else default
+
+    def _temporal_extent_handling(template_conf,collection):
```
kyle-lesinger, line 121: With the newest ingest of California Wildfires, the original data doesn't have any dates in it. This means that pyarc2stac will return `dashboard:is_timeless` = True.

But the issue is that when previewing, it will not have the relevant dates.  If is_timeless is generated by pyarcstac, it would not traditionally be overwritten by `is_periodic`. Therefore, we have to manually overwrite based on the input from the template_conf file.

Since true and false are treated as strings when being based through the `template_conf` manual triggering, this function allows it to actually override the `is_timeless` key/value pair.  After testing, this function properly overrides `is_timeless`.

kyle-lesinger:
@slesaad the conflicts have been resolved. Please re-review at your earliest convenience.

dags/veda_data_pipeline/veda_pyarc2stac_pipeline.py
```diff
@@ -0,0 +1,150 @@
...
+    filtered_template = {k: v for k, v in template_conf.items() if v not in (None, "")}
+    filtered_event = {k: v for k, v in event.items() if v not in (None, "")}
+    
+    # Merge with precedence: template_conf > event > pyarc2stac defaults
+    # Start with collection (pyarc2stac defaults), update with event, then template
+    merged = collection.copy()
+    
+    # Handle temporal extent separately if it exists in configs.
+    # This is useful for items with no temporal extent in the initial pyarc2stac item creation
+    if "temporal" in filtered_event:
```
kyle-lesinger, line 105: For this part, we need to replace the temporal extent for those collections which don't have a date. They will produce a dashboard:is_timeless which overrides the dashboard:time_density. Some of the NASA Disasters data has this issue.

slesaad (approved):

PR #366, opened 2025-05-05
Fix collection group app secret

smohiudd:
Use correct submit_stac handler app secret parameter.

### DIFF
--- a/dags/veda_data_pipeline/groups/collection_group.py
+++ b/dags/veda_data_pipeline/groups/collection_group.py
@@ -37,13 +37,13 @@ def ingest_collection_task(ti=None, collection=None):
         collection = ti.xcom_pull(task_ids='Collection.generate_collection')
     airflow_vars = Variable.get("aws_dags_variables")
     airflow_vars_json = json.loads(airflow_vars)
-    cognito_app_secret = airflow_vars_json.get("COGNITO_APP_SECRET")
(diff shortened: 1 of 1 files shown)


ividito (approved):

anayeaye (approved):

dags/veda_data_pipeline/groups/collection_group.py
```diff
@@ -37,13 +37,13 @@ def ingest_collection_task(ti=None, collection=None):
         collection = ti.xcom_pull(task_ids='Collection.generate_collection')
     airflow_vars = Variable.get("aws_dags_variables")
```
amarouane-ABDELHAK: We can add `deserialize_json` parameter so we will not have to call `json.loads` 
```suggestion
    airflow_vars_json = Variable.get("aws_dags_variables", deserialize_json=True)
```

dags/veda_data_pipeline/groups/collection_group.py
```diff
@@ -37,13 +37,13 @@ def ingest_collection_task(ti=None, collection=None):
...
     airflow_vars_json = json.loads(airflow_vars)
```
amarouane-ABDELHAK:
```suggestion
```

amarouane-ABDELHAK (approved):

PR #368, opened 2025-05-07
Fix/uv for tests

ividito:
Use `uv` for faster tests and CI:

image


amarouane-ABDELHAK (commented):
Looks good. But I think `uv` will create python3.12 virtual environemnet while we are using python 3.11
because of this uv.lock entry
```
resolution-markers = [
    "python_full_version >= '3.13'",
    "python_full_version == '3.12.*'",
    "python_full_version < '3.12'",
]
```

amarouane-ABDELHAK (approved):

PR #369, opened 2025-05-12
Add sm2a test for veda deploy

amarouane-ABDELHAK:

Run SM2A unit tests before deploying SM2A via veda-deploy 
## Changes

*Add running unit tests to terraform-deploy-sm2a

### DIFF
--- a/.github/actions/terraform-deploy-sm2a/action.yml
+++ b/.github/actions/terraform-deploy-sm2a/action.yml
@@ -31,7 +31,6 @@ runs:
       uses: astral-sh/setup-uv@v5
       with:
         python-version: "3.11"
-        cache: "pip"
   
     - name: Setup Terraform
       uses: hashicorp/setup-terraform@ed3a0531877aca392eb870f440d9ae7aba83a6bd #v1.4.0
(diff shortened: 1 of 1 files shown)


.github/actions/terraform-deploy-sm2a/action.yml
```diff
@@ -28,23 +28,33 @@ runs:
...
+
     - name: Deploy
       working-directory: ${{ inputs.dir }}
       shell: bash
       env:
         AWS_DEFAULT_REGION: ${{ inputs.aws-region }}
         AWS_REGION: ${{ inputs.aws-region }}
       run: |
+        # Activate uv venv
+        source ../.venv/bin/activate
```
anayeaye, line 57: Does this venv already exist on the runner?

amarouane-ABDELHAK: Yes, it will be created during `uv` setup

anayeaye: Cool. This uv stuff is new to me. Is there reason to deactivate at the end of the job in case another job follows on the same runner? Informational question only, not change request!

amarouane-ABDELHAK: It automatically deactivates when the step is complete. You'll need to manually activate it for the next step. It's an annoying thing of GitHub workflows but in this use case it actually turns out to be a useful feature, lol the irony.

ividito: `uv run make ...` should also work here, but this is more explicit anyway

PR #373, opened 2025-05-13
Bring back scheduled DAGs

### LINKED ISSUE #372: Scheduled DAGs no longer generated
## What
DAGs are not being generated for the configs in the event bucket. Possibly related to DAG deduplication added in PR #367 

## AC
- [ ] DAGs are generated for all configs in tf-event-bucket/collections

ividito:
**Summary:** 

Fixes #372 
Currently live on https://sm2a.sit.openveda.cloud

Accessing the DagBag seems to short-circuit generating new DAGs with `generate_dags`. 

Airflow's `dag_bag.py` only performs a top-level scan of the files it parses. Because `generate_dags.py` uses functions from `veda_discover_pipeline.py`, as far as the `DagBag` is concerned, `veda_discover_pipeline` generates the same DAGs as `generate_dags.py` (this was hard to debug, but this is the best explanation I could come up with)

Duplicate DAGs are already blocked by Airflow - these changes are to satisfy the integrity constraints in our unit tests. In the future, we could consider removing these unit tests and adding integration tests against the `docker compose` stack, in order to validate DAG integrity after DAGs have been fully processed.

## PR Checklist

- [x] Unit tests
- [ ] Integration tests

### DIFF
--- a/dags/generate_dags.py
+++ b/dags/generate_dags.py
@@ -10,7 +10,7 @@ from veda_data_pipeline.veda_discover_pipeline import get_discover_dag
 from veda_data_pipeline.veda_vector_pipeline import get_ingest_vector_dag
 
 from airflow.models import DagBag
-existing_dag_ids = set(DagBag().dags.keys())
+existing_dag_ids = DagBag().dag_ids
 
 def filter_configs_by_dag(
         collection_configs: List[Dict[str, int]],
--- a/dags/generate_dags.py
+++ b/dags/generate_dags.py
@@ -9,9 +9,6 @@ from typing import Dict, List, Optional
 from veda_data_pipeline.veda_discover_pipeline import get_discover_dag
 from veda_data_pipeline.veda_vector_pipeline import get_ingest_vector_dag
 
-from airflow.models import DagBag
-existing_dag_ids = DagBag().dag_ids
-
 def filter_configs_by_dag(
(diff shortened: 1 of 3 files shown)


dags/generate_dags.py
```diff
@@ -90,12 +84,10 @@ def generate_dags():
...
-                print(f"Skipping duplicate DAG ID: {id}")
-                continue
             get_ingest_vector_dag(
                 id=id, event=vector_config
             )
 
-
 generate_dags()
+get_ingest_vector_dag(id="veda_ingest_vector", event={})
+get_discover_dag(id="veda_discover", event={})
```
anayeaye, line 93: Probably just a style decision but because the other unscheduled DAGs are cerated in the *_pipeline.py files, I wonder if we should keep that convention and leave the generate_dags.py file for just creating the scheduled DAGs?

anayeaye (approved):
One organization question about generating the non scheduled DAGs consistently. The SIT deployment looks good as do the test in the PR action.

dags/generate_dags.py
ividito: I would prefer this too, but it breaks the `DagBag`-based tests. I left the method calls in the original files commented out, since I think we should switch back after revisiting our DAG integrity testing strategy.

PR #376, opened 2025-05-20
Enforce define slack notification for all dags

amarouane-ABDELHAK:

Enforce defining slack notification for all dags, so we will get notification in case of failures
## Changes

* Added unit tests to core dags test


ividito (changes requested):
Could we consolidate our default DAG parameters to reduce duplication? Options like `tags`, `doc_md` or `params` should be per-DAG, but other parameters we're defaulting could be set in one place (like the failure callback).

The other benefit to de-duplicating here is that we could insert a different, less noisy callback in non-production environments, without needing to duplicate it in each DAG or include an environment check in the `slack_fail_alert`.

ividito (approved):

Issue #370, opened 2025-05-12
Use dataset events to store ingestion config data


smohiudd:
## Description

See [Ingestion Config Persistence ADR](https://github.com/NASA-IMPACT/veda-architecture/blob/main/adrs/20250326-ingestion-config-persistence.md).

PR #318, opened 2025-03-13
Update Airflow to 2.10.5

ividito:
**Summary:** 

Airflow (+related libraries) version upgrade. This prompted some other changes and improvements, as several things broke with the updated libraries.
- old SQS image is outdated and had to be replaced
- switched to uv and dropped conda to fix library build issues
- example DAGs use outdated providers, I opted to delete instead of refactor
- `LazyXcomAccess` -> `LazyXcomSequence`
- stactools DAGs were scheduled by default - this has been fixed

~Manual step required to make deployment work with new version - `python scripts/run_task.py --vpc-id vpc-xxx --security-group xxx --task-definition xxx --cluster xxx--command 'db init'`~

^ No longer applies, instead we must set the `airflow_version` variable for SM2A - https://github.com/NASA-IMPACT/self-managed-apache-airflow/pull/23

## Testing/action items
- [x] SIT deployment tested (deploy works, but services require manual db upgrade)
- [x] tested raster ingestion with `veda-dataset-pipeline`
- [x] tested vector ingestion with `veda-vector-pipeline` - works, but SIT cannot connect to dev features DB to submit generated items

PR #383, opened 2025-06-23
Trace dataset configs using Airflow Datasets/Assets

ividito:
**Summary:**

Addresses #370 

## Changes

- This PR is based on #318 in order to leverage Dataset Aliases.
- All ingests using the `veda-datasets` DAG will produce a Dataset event, which is correlated by collection ID.
- These events also write a manifest to s3 under the `event_bucket`
- In Airflow 3, the "Datasets" terminology changes to "Assets" - this is no less confusing for us, but explains the deprecation warnings introduced by this PR. We can revisit this if/when we upgrade to Airflow 3.

## PR Checklist

- [ ] Unit tests
- [x] Ad-hoc testing - Deploy changes and test manually (currently live on SIT)

### DIFF
--- a/dags/veda_data_pipeline/groups/processing_tasks.py
+++ b/dags/veda_data_pipeline/groups/processing_tasks.py
@@ -4,6 +4,7 @@ import logging
 from copy import deepcopy
 import smart_open
 from airflow.models.variable import Variable
+from airflow.models.xcom import LazyXComSelectSequence
 from airflow.decorators import task
 from airflow.datasets import Dataset, DatasetAlias
 from airflow.datasets.metadata import Metadata
@@ -88,7 +89,7 @@ def build_stac_task(payload, ti=None):
             DatasetAlias("VEDA-Datasets")
         ],
 )
-def post_ingest_report(ti, logical_date):  # params are Airflow kwargs - use this task without input
+def post_ingest_dataset_event(ti, logical_date, built_items = {}):  # params are Airflow kwargs - use this task without input
     """
     Logs a Dataset event, saving the config used as a versioned object in s3, and creating a Metadata object visible in Airflow.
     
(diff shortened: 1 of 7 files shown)


dags/veda_data_pipeline/groups/processing_tasks.py
```diff
@@ -80,3 +83,53 @@ def build_stac_task(payload, ti=None):
...
+        Yields a Metadata object that Airflow uses to register the Dataset event.
+    """
+    payload = ti.dag_run.conf
+    event_bucket_name = Variable.get("aws_dags_variables", deserialize_json=True).get("EVENT_BUCKET")
+    collection = payload.get("collection", None)
+    if not collection:
+        raise ValueError("Collection ID is required in the payload to create a report.")
+    
+    # write the payload to S3 as a versioned object
+    key = f"s3://{event_bucket_name}/airflow_events/{collection}/{logical_date.isoformat()}.json"
```
aliziel: Clarifying question since I'm still getting up to speed on decisions: are these ever going to be accessed programmatically ? Asking because the special characters in the object key might add some headache down the line, e.g. `s3://bucket/airflow_events/collection/2025-06-23T14:44:20+00:00.json`. Looks like they might be used to associate records, so perhaps necessary !

scripts/deploy.sh
```diff
@@ -91,7 +91,7 @@ check_create_remote_state $AWS_REGION $STATE_BUCKET_NAME $STATE_DYNAMO_TABLE
 read -rp 'action [init|plan|deploy]: ' ACTION
 case $ACTION in
   init)
-    terraform init
+    terraform init -reconfigure
```
botanical:
```suggestion
    terraform init
init-reconfigure)
    terraform init -reconfigure
```

Could we instead have an option to reconfigure? What would be the benefit to always include reconfiguring on the init? 😯

dags/veda_data_pipeline/groups/processing_tasks.py
```diff
@@ -80,3 +83,53 @@ def build_stac_task(payload, ti=None):
...
+    with smart_open.open(key, "w") as f:
+        json.dump(payload, f, indent=2)
+    log_task(f"Payload written to {key}")
```
botanical: should we put this in a try/catch with error handling or nah?

botanical (commented):
Had a couple of small comments. Overall, I think using Airflow Datasets to trace our dataset configs is such a good idea 💡 👏 Thanks for this PR!

scripts/deploy.sh
ividito: This was a solution to a specific issue I was having on SIT - basically brute-force accepting the current deployed state. Definitely not needed.

dags/veda_data_pipeline/groups/processing_tasks.py
ividito: Changed the formatting to strip special characters while still keeping the resulting string sortable.

botanical (approved):

PR #391, opened 2025-08-01
update regex pattern

kyle-lesinger:
**Summary:**

Update regex pattern matching in `build_stac/utils/regex.py` to be more inclusive for dates that begin at the start of the filename. 

E.g., 20240914_DSWx-S1_BWTR.tif or 20240926_DSWx-S1_WTR.tif. Currently airflow will return an error stating that there is no valid datetime string found in the file.

## Changes

* Update regex to allow for dates to be found if they begin at the start of a string.

### DIFF
--- a/dags/veda_data_pipeline/utils/build_stac/utils/regex.py
+++ b/dags/veda_data_pipeline/utils/build_stac/utils/regex.py
@@ -4,6 +4,7 @@ from typing import Callable, Dict, Tuple, Union
 
 from dateutil.relativedelta import relativedelta
 
+from . import events
 
 DATERANGE = Tuple[datetime, datetime]
 
--- a/dags/veda_data_pipeline/utils/build_stac/utils/regex.py
+++ b/dags/veda_data_pipeline/utils/build_stac/utils/regex.py
@@ -26,7 +26,7 @@ def _calculate_day_range(datetime_obj: datetime) -> DATERANGE:
     return start_datetime, end_datetime
 
 
-DATETIME_RANGE_METHODS: Dict[INTERVAL, Callable[[datetime], DATERANGE]] = {
+DATETIME_RANGE_METHODS: Dict[events.INTERVAL, Callable[[datetime], DATERANGE]] = {
     "month": _calculate_month_range,
     "year": _calculate_year_range,
     "day": _calculate_day_range,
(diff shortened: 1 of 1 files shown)


dags/veda_data_pipeline/utils/build_stac/utils/regex.py
```diff
@@ -27,27 +26,27 @@ def _calculate_day_range(datetime_obj: datetime) -> DATERANGE:
...
-        (r"[_\.\-](\d{8})", "%Y%m%d"),
-        (r"[_\.\-](\d{6})", "%Y%m"),
-        (r"[_\.\-](\d{4})", "%Y"),
+        (r"(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})", "%Y-%m-%dT%H:%M:%S"),
+        (r"(\d{8}T\d{6})", "%Y%m%dT%H%M%S"),
+        (r"(\d{4}_\d{2}_\d{2})", "%Y_%m_%d"),
+        (r"(\d{4}-\d{2}-\d{2})", "%Y-%m-%d"),
+        (r"(\d{8})", "%Y%m%d"),
+        (r"(\d{6})", "%Y%m"),
+        (r"(\d{4})", "%Y"),
```
slesaad, line 50: Only keep these changes and revert the others. This change seems reasonable.

kyle-lesinger: @anayeaye there were some files for NASA Disasters that have the date string as the very beginning of the file (e.g., 20241029_<other info>.tif). I will be giving a short demo for some of the data that I've ingested for them tomorrow and I will be making several suggestions to them regarding file format and naming conventions. 

It may be possible that they will need to re-process their data so I could make suggestions that these filenames should be changed to have the date at a later part of the string (e.g., at the end).

Do you foresee that changing that this PR's regex change to allow for dates at the beginning of a string will have any major issues?

kyle-lesinger: > Only keep these changes and revert the others. This change seems reasonable.

Other changes have been reverted.

kyle-lesinger:
Closing. We have worked with Disasters staff to rename file in accordance with regex patterns for ingestion. We will be renaming files.

PR #403, opened 2025-09-24
Change the public subnet filter logic

paridhi-parajuli:
Changes:
1. Change the public subnet filter logic as a fix to [this issue](https://github.com/NASA-IMPACT/veda-deploy/actions/runs/18019792476/job/51275885297)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ test = [
     "stac_pydantic",
     "mypy_boto3_s3",
     "rasterio",
-    "rio_stac==0.12.0",
+    "rio_stac",
     "virtualenv",
 ]
 airflow-worker = [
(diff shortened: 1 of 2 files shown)


slesaad (approved):

pyproject.toml
```diff
@@ -58,7 +58,7 @@ airflow-worker = [
     "xstac",
     "pystac",
     "rasterio",
-    "rio-stac",
+    "rio_stac",
```
anayeaye, line 61:
```suggestion
    "rio-stac",
```
I think this should be a hyphen not an underscore?

anayeaye: Sorry, I was wrong about that. Looks like underscore is right.

anayeaye (approved):

anayeaye:
I reviewed the subnets in some of our existing stacks and verified that the new selection will catch them as well

Issue #402, opened 2025-09-15
Upgrade pystac to v1.14.0


anayeaye:
## What
Upgrade pystac to resolve `urllib` import error.

Refs: 
- https://github.com/OSGeo/PROJ/issues/4567
- https://github.com/stac-utils/pystac/releases/tag/v1.14.0

## AC
- [ ] Pystac upgraded to v1.14.0

PR #408, opened 2025-10-22
upgrade pystack to v1.14.0

sandrahoang686:
Ticket: https://github.com/NASA-IMPACT/veda-data-airflow/issues/402


sandrahoang686:
Testing:
1. How do i replicate the original bug error reported from https://github.com/OSGeo/PROJ/issues/4567 ?
2. How do check if this change resolves that error and installs the correct version?

PR #409, opened 2025-10-27
Update pystac version to 1.14.1

sandrahoang686:
**Summary:**

Addresses https://github.com/NASA-IMPACT/veda-data-airflow/issues/402

[v1.14.1](https://github.com/stac-utils/pystac/blob/main/CHANGELOG.md#v1141---2025-09-18) supports older versions of urllib3


sandrahoang686:
I saw that Pystac was only pinned in those two files that i've updated in this PR. So I've bumped the versions there but am i missing anywere else?

Also I wasn't able to replicate (not too sure how) the original error reported [here](https://github.com/OSGeo/PROJ/issues/4567). So unsure if this really resolves it. Is there a way I could test this? I've just tried having the services run locally on 8080.

cc @anayeaye

airflow_worker/Dockerfile
```diff
@@ -44,6 +44,6 @@ RUN cp ${AIRFLOW_HOME}/configuration/airflow.cfg* ${AIRFLOW_HOME}/.
 ENV AIRFLOW_HOME ${AIRFLOW_HOME}
 ENV TZ UTC
 ENV PYTHONPATH /opt/airflow
-ENV PYSTAC_STAC_VERSION_OVERRIDE=1.0.0
+ENV PYSTAC_STAC_VERSION_OVERRIDE=1.14.1
```
anayeaye:
```suggestion
ENV PYSTAC_STAC_VERSION_OVERRIDE=1.0.0
```
this one is actually setting pystac to check that the stac metadata is stac version 1.0.0 so we want to keep it

anayeaye (approved):
Approve after one change is applied

dags/veda_data_pipeline/utils/cogify_transfer/requirements.txt
```diff
@@ -1,7 +1,7 @@
 aws-lambda-powertools
 awslambdaric
 boto3
-pystac==1.4.0
+pystac==1.14.1
```
anayeaye:
```suggestion
pystac
```
we probably don't even need to pin the version here (we are not pinning in other locations)

anayeaye:
@sandrahoang686 just a small change (stac version) and then I suggest we don't pin pystac version anymore, either.

PR #411, opened 2025-11-20
Test - DAG for OCO3-SAM data

siddharth0248:
**Summary:**

Create a DAG that fetches only new files from external S3 bucket and copy it to ghg-data-store. 
Next step would be to ingest the new items
https://github.com/US-GHG-Center/ghgc-data/pull/65


siddharth0248:
@smohiudd @anayeaye

PR #414, opened 2025-12-11
feat: tenant tagging pipeline draft

botanical:
**Summary:**

https://github.com/NASA-IMPACT/veda-architecture/issues/677

## Changes

* Created a DAG to tag a list of collections with a tenant property
* Added a function to fix normalizing dates, which I noticed when I was testing updating collections with my DAG

## PR Checklist

- [ ] Unit tests
- [ ] Integration tests
- [ ] Infrastructure changes - test using `terraform validate` and `terraform plan`

### DIFF
--- a/dags/veda_data_pipeline/utils/collection_generation.py
+++ b/dags/veda_data_pipeline/utils/collection_generation.py
@@ -3,7 +3,7 @@ from typing import Any, Dict
 import fsspec
 import xarray as xr
 import xstac
-from veda_data_pipeline.utils.schemas import SpatioTemporalExtent, normalize_datetime_to_iso8601
+from veda_data_pipeline.utils.schemas import SpatioTemporalExtent
 from datetime import datetime, timezone
 
 
@@ -101,13 +101,14 @@ class GenerateCollection:
             collection_stac["extent"]["spatial"] = {"bbox": [list(spatial_extent.values())]}
 
         if temporal_extent := dataset.get("temporal_extent"):
-            # Normalize temporal extent values to ISO 8601 format
-            normalized_values = [
-                normalize_datetime_to_iso8601(x) for x in list(temporal_extent.values())
-            ]
-
             collection_stac["extent"]["temporal"] = {
-                "interval": [normalized_values]
+                "interval": [
+                    [
+                        x
+                        if x else None
+                        for x in list(temporal_extent.values())
+                    ]
+                ]
             }
 
         collection_stac["item_assets"] = {
--- a/dags/veda_data_pipeline/utils/collection_generation.py
+++ b/dags/veda_data_pipeline/utils/collection_generation.py
@@ -3,7 +3,7 @@ from typing import Any, Dict
 import fsspec
 import xarray as xr
 import xstac
-from veda_data_pipeline.utils.schemas import SpatioTemporalExtent
+from veda_data_pipeline.utils.schemas import SpatioTemporalExtent, normalize_datetime_to_iso8601
 from datetime import datetime, timezone
 
 
@@ -99,16 +99,15 @@ class GenerateCollection:
         # Override the extents if they exists
         if spatial_extent := dataset.get("spatial_extent"):
             collection_stac["extent"]["spatial"] = {"bbox": [list(spatial_extent.values())]}
-        
+
         if temporal_extent := dataset.get("temporal_extent"):
+            # Normalize temporal extent values to ISO 8601 format
+            normalized_values = [
+                normalize_datetime_to_iso8601(x) for x in list(temporal_extent.values())
+            ]
+
             collection_stac["extent"]["temporal"] = {
-                "interval": [
-                    [
-                        x
-                        if x else None
-                        for x in list(temporal_extent.values())
-                    ]
-                ]
+                "interval": [normalized_values]
             }
 
         collection_stac["item_assets"] = {
--- a/dags/veda_data_pipeline/utils/schemas.py
+++ b/dags/veda_data_pipeline/utils/schemas.py
@@ -64,15 +64,11 @@ def normalize_temporal_extent(collection: Dict[str, Any]) -> Dict[str, Any]:
     if "extent" in collection and "temporal" in collection["extent"]:
         temporal = collection["extent"]["temporal"]
         if "interval" in temporal and isinstance(temporal["interval"], list):
-            normalized_intervals = []
-            for interval in temporal["interval"]:
-                if isinstance(interval, list):
-                    normalized_interval = [
-                        normalize_datetime_to_iso8601(dt) for dt in interval
-                    ]
-                    normalized_intervals.append(normalized_interval)
-                else:
-                    normalized_intervals.append(interval)
-            collection["extent"]["temporal"]["interval"] = normalized_intervals
+            collection["extent"]["temporal"]["interval"] = [
+                [normalize_datetime_to_iso8601(dt) for dt in interval]
+                if isinstance(interval, list)
+                else interval
+                for interval in temporal["interval"]
+            ]
 
     return collection
--- a/dags/veda_data_pipeline/utils/schemas.py
+++ b/dags/veda_data_pipeline/utils/schemas.py
@@ -1,8 +1,10 @@
 # Description: Lightweight schema definitions
 
 from datetime import datetime
-from typing import List, Union
+from typing import List, Union, Any, Dict
 from stac_pydantic.collection import Extent, TimeInterval
+from pystac.utils import datetime_to_str
+from dateutil import parser as date_parser
 
 
 class DatetimeInterval(TimeInterval):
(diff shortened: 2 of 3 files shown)


dags/veda_data_pipeline/utils/collection_generation.py
```diff
@@ -99,16 +99,15 @@ def create_cog_collection(self, dataset: Dict[str, Any]) -> dict:
...
+
             collection_stac["extent"]["temporal"] = {
-                "interval": [
-                    [
-                        x
-                        if x else None
-                        for x in list(temporal_extent.values())
-                    ]
-                ]
+                "interval": [normalized_values]
```
botanical: @smohiudd do you know if this is where the existing date bug is?

smohiudd: I think it may be in the ingest API

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,297 @@
...
+        default=None,
+        type=["null", "array"],
+        description="List of collection IDs to tag (optional if catalog_endpoint is provided)"
+    ),
+    "catalog_endpoint": Param(
+        default=None,
+        type=["null", "string"],
+        description="STAC catalog endpoint URL to fetch all collections from (optional if collections is provided)"
+    ),
+    "tenant": Param(default=None, type="string", description="Tenant ID to tag the collection with (will be set as eic-tenant property)"),
```
anayeaye: Because tenant filter field is configurable in the backend should we use two keys in this config `tenant_filter_field` and `tenant_filter_value`?
https://github.com/NASA-IMPACT/veda-backend/blob/dbd05924d7c04eeeec9ed88cc445a7b33a312261/stac_api/runtime/src/config.py#L103

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,297 @@
...
+
+logger = logging.getLogger(__name__)
+
+template_dag_run_conf = {
+    "collections": Param(
+        default=None,
+        type=["null", "array"],
+        description="List of collection IDs to tag (optional if catalog_endpoint is provided)"
+    ),
+    "catalog_endpoint": Param(
```
anayeaye: Discussion: I would almost rather have this DAG accept an array of collection ids so that we can use it on a subset of the collections in a catalog. I'm still reading through the methods below for adding a tag to all existing collections in a catalog to think it over.
cc: @botanical @smohiudd

botanical: Okay, let me know! It takes in either a catalog endpoint or a list of collection IDs so I can just modify it to remove the catalog endpoint parameter @anayeaye

anayeaye: Oh cool--I saw the catalog first. The collection ids make sense to me and let's chat about pros/cons on updating an entire catalog. I'll resolve this and we can chat in stand up

botanical: Talked offline, we're going to probably remove the collections catalog endpoint parameter since it's an unlikely scenario where we want to set a whole catalog to a single tenant in multi-tenancy

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,354 @@
...
+            )
+            logger.error(error_msg)
+            raise ValueError(error_msg)
+        tenant_field = tenant_field.strip()
+
+        logger.debug(f"Existing collection keys: {list(existing_collection.keys())}")
+
+        updated_collection = existing_collection.copy()
+
+        if "properties" not in updated_collection:
```
anayeaye: Properties are not part of the spec for collections so we shouldn't require. I'm don't know that we ever would want to add a properties object to a collection. 🤔

botanical: https://github.com/NASA-IMPACT/veda-data-airflow/pull/414/commits/1c384ca6d006c31cf9726a10f9d18e98f4da63d6

dags/veda_data_pipeline/utils/schemas.py
```diff
@@ -13,3 +15,64 @@ class DatetimeInterval(TimeInterval):
...
+            normalized_intervals = []
+            for interval in temporal["interval"]:
+                if isinstance(interval, list):
+                    normalized_interval = [
+                        normalize_datetime_to_iso8601(dt) for dt in interval
+                    ]
+                    normalized_intervals.append(normalized_interval)
+                else:
+                    normalized_intervals.append(interval)
+            collection["extent"]["temporal"]["interval"] = normalized_intervals
```
sandrahoang686: ⛏️ 
```suggestion
            collection["extent"]["temporal"]["interval"] = [
                [normalize_datetime_to_iso8601(dt) for dt in interval]
                if isinstance(interval, list)
                else interval
                for interval in temporal["interval"]
            ]
```
? maybe

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,327 @@
...
+
+        if not collections:
+            error_msg = "The 'collections' list must be provided in DAG configuration"
+            logger.error(error_msg)
+            raise ValueError(error_msg)
+
+        if not isinstance(collections, list):
+            error_msg = f"Collections must be a list, but got type: {type(collections)}"
+            logger.error(error_msg)
+            raise ValueError(error_msg)
```
sandrahoang686, line 94: I'm just running through these cases and I was wondering - so are we expecting at line 86 to get a string like so `"col1, col2, col3"`? If so and line 87 proceeds sucessfully then does condition 92 and 97 ever get reached? should 92 be checking for `collections_strings` instead? and then 97 might not be needed because it wont reach - if `split()` which should always return an array doesn't succeed, it errors out right away and proceeds to the exception I think? but not sure

dags/veda_data_pipeline/utils/schemas.py
botanical: We could do it your suggested way and use list comprehension! I think it's functionally the same

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
botanical: Thanks for all the questions @sandrahoang686! I think you're right that line 997 might not be reachable 👀 since line 87, with `split()` will return a list. Let me simplify this so that the pipeline doesn't try to validate and parse a list of lists as input, and instead only accepts a list of strings

botanical: https://github.com/NASA-IMPACT/veda-data-airflow/pull/414/changes/80f75d62fb57314bfe4b3249c033f7d6e9a80a76

botanical:
@sandrahoang686 Thanks for the review, I believe I addressed all your feedback. 

I made one additional change to update the DAG to handle newlines for a collections list since this seems to be the convention that airflow uses for list params. 
https://airflow.apache.org/docs/apache-airflow/stable/_modules/airflow/example_dags/example_params_ui_tutorial.html
https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html#use-params-to-provide-a-trigger-ui-form

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,348 @@
...
+with DAG("veda_tenant_tagging_pipeline", params=template_dag_run_conf, **dag_args) as dag:
+    start = EmptyOperator(task_id="start", dag=dag)
+    end = EmptyOperator(task_id="end", dag=dag)
+
+    collection_ids = get_collection_ids()
+    fetch_collections = fetch_existing_collection.expand(collection_id=collection_ids)
+    update_collections = update_collection_with_tenant_tags.expand(existing_collection=fetch_collections)
+    ingest_collections = ingest_all_collections(collections=update_collections)
+
+    start >> collection_ids >> fetch_collections >> update_collections >> ingest_collections >> end
```
ividito: Style change to only define DAG edges in one place (just a preference)

```suggestion
    collection_ids = start >> get_collection_ids()
    fetch_collections = fetch_existing_collection.expand(collection_id=collection_ids)
    update_collections = update_collection_with_tenant_tags.expand(existing_collection=fetch_collections)
    ingest_collections = ingest_all_collections(collections=update_collections) >> end
```

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,348 @@
...
+from slack_notifications import slack_fail_alert
+import requests
+from airflow.models.variable import Variable
+
+logger = logging.getLogger(__name__)
+
+template_dag_run_conf = {
+    "collections": Param(
+        default=None,
+        type=["null", "array"],
```
ividito: Is "null" ever applicable here? If we have this as just "array" (making it required), does it still work to only input one collection ID in the UI? Apologies if this is a duplicate of @sandrahoang686's earlier questions

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
```diff
@@ -0,0 +1,348 @@
...
+            results.append({"collection_id": collection_id, "status": "error", "error": error_msg, "payload_size_bytes": payload_size})
+            # Continue processing other collections instead of failing immediately
+            continue
+
+        # I put this small delay between requests in case we need to avoid rate limiting
+        if idx < total:
+            time.sleep(0.5)
+
+    successful = sum(1 for r in results if r.get("status") == "success")
+    failed = sum(1 for r in results if r.get("status") == "error")
```
ividito: nit (`True` evaluates to 1 for `sum`)
```suggestion
    successful = sum(r.get("status") == "success" for r in results)
    failed = sum(r.get("status") == "error" for r in results)
```

ividito (approved):
Some nits but no blockers

dags/veda_data_pipeline/veda_tenant_tagging_pipeline.py
botanical: Ah thanks for catching that! Yeah, null is not applicable here. It does still work if we input one collection ID but it needs to be an array (so a single element in an array).

botanical: https://github.com/NASA-IMPACT/veda-data-airflow/pull/414/commits/b24a824bd67a5f87e743e0c09824a9db6b231516

PR #418, opened 2026-01-23
feat: alternate FAB for auth with veda-keycloak

ividito:
This PR will require additional envs

```
TF_VAR_keycloak_base_url=https://keycloak.delta-backend.xyz
TF_VAR_keycloak_realm=veda
TF_VAR_keycloak_client_id=airflow-webserver-fab
TF_VAR_keycloak_client_secret=<get this from keycloak>
```

Tested with https://sm2a.sit.openveda.cloud/login/


airflow_services/webserver_config_keycloak.py
```diff
@@ -0,0 +1,216 @@
...
+    
+    Adjust this function based on your Keycloak configuration.
+    """
+    roles = []
+
+    log.info(f"Extracting roles from Keycloak response: {resp}")
+    access_token = resp.get("access_token", "")
+    
+    try:
+        decoded = jwt.decode(access_token, options={"verify_signature": False})
```
botanical: I think this is the right approach but I think it would be better if we made sure the access token is only from keycloak if that makes sense. So I would suggest using this introspection endpoint they offer https://www.keycloak.org/securing-apps/oidc-layers#_token_introspection_endpoint and use the token info given from the introspection endpoint to extract the roles

botanical: @ividito  I did some more research about login workflows using keycloak for airflow, and I think I’m wrong about the introspection endpoint kind of. It’s still useful but we could also make it more secure and avoid the additional network call by verifying the signature using keycloak’s public key https://airflow.apache.org/docs/apache-airflow-providers-fab/stable/auth-manager/webserver-authentication.html
```
# Fetch public key
req = requests.get(OIDC_ISSUER)
key_der_base64 = req.json()["public_key"]
key_der = b64decode(key_der_base64.encode())
public_key = serialization.load_der_public_key(key_der)


class CustomSecurityManager(FabAirflowSecurityManagerOverride):
    def get_oauth_user_info(self, provider, response):
        if provider == "keycloak":
            token = response["access_token"]
            me = jwt.decode(token, public_key, algorithms=["HS256", "RS256"], audience="account")
```
^ particularly this bit I think is helpful.

smohiudd:
I have this branch deployed to SIT: https://sm2a.sit.openveda.cloud/

smohiudd (approved):

PR #433, opened 2026-03-30
Disasters/add STAC metadata

kyle-lesinger:
**Summary:**

Addresses [VEDA-XX: Develop amazing new feature](https://github.com/NASA-IMPACT/veda-data-pipelines/issues/XX)

## Changes

* Detailed list or prose of changes
* ...

## PR Checklist

- [ ] Unit tests
- [ ] Integration tests
- [ ] Infrastructure changes - test using `terraform validate` and `terraform plan`


kyle-lesinger:
@anayeaye @smohiudd this branch will also be useful in the future for testing with any new additions to the STAC schema (e.g., EGIS service href). This was suggested previously to expose the service within the STAC browser natively. Any thoughts?

aws_dags_variables_instructions.md
```diff
@@ -0,0 +1,81 @@
+# AWS DAGs Variables Configuration
```
anayeaye, line 1: the aws_dags_variables in secrets manager is generated during the deployment of self managed airflow using values from the deployment secret. We do not want to edit aws_dags_variables directly but I also am admittedly having a hard time finding the definitions for the deployment envs that we can and should modify (https://github.com/NASA-IMPACT/self-managed-apache-airflow/blob/e67e48baad16674e5f14a14cbf55290c36304ccd/infrastructure/secrets/main.tf#L56)

I found some base docs in veda-deploy but we need more clear definitions of how to update deployment envs (I think we lost them in a refactor) https://github.com/NASA-IMPACT/veda-deploy?tab=readme-ov-file#aws-secrets-requirements-for-sm2a

Let's work to improve the deployment env definitions with some of the definitions you have here but not overwriting the terraformed secret.

aws_dags_variables_instructions.md
```diff
@@ -0,0 +1,81 @@
...
+
+This JSON file contains the configuration variables that Airflow needs to run the VEDA data pipelines.
+
+## How to Use
+
+### Option 1: Upload to AWS Secrets Manager (Recommended for Production)
+```bash
+aws secretsmanager create-secret \
+  --name veda-pipeline-dev/airflow/variables/aws_dags_variables \
+  --secret-string file://aws_dags_variables.json \
```
ividito, line 11: This flag will not work like this - it doesn't unpack the file contents into valid json

dags/veda_data_pipeline/utils/disasters_utils.py
```diff
@@ -0,0 +1,310 @@
...
+def _load_lut(filename: str) -> dict:
+    """Load JSON data from the local LUT directory."""
+    with open(os.path.join(_LUT_DIR, filename)) as f:
+        return json.load(f)
+
+
+@lru_cache(maxsize=1)
+def _get_country_codes_mapping() -> Dict[str, List[str]]:
+    """Load and cache country codes mapping from S3."""
+    data = _load_lut("monty_country_codes.json")
```
ividito, line 20: I think there's a risk with having LUTs stored as local-to-the-worker files - any changes would have to go through the PR -> review -> merge -> deploy process. It might fit the use case better for these to be managed as s3 files with Airflow Variables as pointers. Something like this: 

```
{
    "event_hazard_location": "s3://veda-tf-state-shared-disasters/reference/event-hazard-location.json",
    "country_codes": "s3://veda-tf-state-shared-disasters/reference/monty_country_codes.json",
    "hazard_codes": "s3://veda-tf-state-shared-disasters/reference/monty_hazard_codes.json"
  }

  lut_config = Variable.get("disasters_lut_paths", deserialize_json=True)
  event_mapping = _load_lut_from_s3(lut_config["event_hazard_location"])
  ...
```

Changes could be made in s3 and automatically applied to new tasks.

dags/veda_data_pipeline/utils/disasters_utils.py
```diff
@@ -0,0 +1,310 @@
...
+    if not event:
+        return {"monty:hazard_codes": None}
+    parsed = _parse_event_string(event)
+    if not parsed:
+        return {"monty:hazard_codes": None}
+    hazard = _get_hazard_codes_mapping().get(parsed["hazard_type"])
+    return {"monty:hazard_codes": [hazard["glide_code"], hazard["classification_code"]] if hazard else None}
+
+
+def extract_hazard_and_location_from_geotiff(file_path: str) -> dict:
```
ividito, line 117: This and other functions aren't called anywhere - we should remove them if there isn't a plan for them
```
  extract_country_codes_from_geotiff
  extract_hazard_codes_from_geotiff
  extract_hazard_and_location_from_geotiff
  extract_corr_id_from_geotiff
```

dags/veda_data_pipeline/utils/s3_discovery.py
```diff
@@ -75,8 +77,14 @@ def discover_from_s3(
                 yield s3_object
 
 
-def group_by_item(discovered_files: List[str], id_regex: str, assets: dict) -> dict:
-    """Group assets by matching regex patterns against discovered files."""
+def group_by_item(discovered_files: List[str], id_regex: str, assets: dict, extract_event_name: bool = False, extract_monty: bool = False, add_product: bool = False, add_providers: bool = False) -> dict:
```
ividito, line 80: There's a lot of disasters-specific flags here (and in other signatures in this file). Would it make more sense to have these moved to another task that looks like this:

```
@task
  def enrich_disaster_metadata(discovery_result: dict, extractors: list[str]) -> dict:
      for item in discovery_result["objects"]:
          file_path = next(iter(item["assets"].values()))["href"]
          if "monty" in extractors:
              item["properties"].update(extract_all_metadata_from_geotiff(file_path))
          # ...
      return discovery_result
```

This post-processing task could be included in a disasters-specific DAG, or could be a future home for other post-discovery metadata enrichment.

dags/veda_data_pipeline/utils/disasters_utils.py
anayeaye: I actually asked to move these tables into git in order to transport them between instances while testing and to move them _out_ of the tf state shared bucket because I didn't think it made sense to re-use that bucket. However, I just had another look and I see that we are already reusing it for scheduling collection discovery events so that ship has sailed. So perhaps formalizing a reference directory in the tf shared state bucket does make sense.

I do think we need to provide documentation for where those files should be stored and probably make an informative error message for when the intended files are not found so we don't think we are having an iam/bucket permissions issue.

dags/veda_data_pipeline/utils/s3_discovery.py
anayeaye: 💡

PR #447, opened 2026-04-22
security: prevent arbitrary table deletion in veda_vector_delete DAG

Adraca:
The veda_vector_delete.py DAG allows specifying a collection_name via dag_run.conf. This is used in a DROP TABLE ... CASCADE statement. While sql.Identifier is used, the lack of value-level validation allows an authorized user to delete any table in the database. Verified via Docker PoC (Node 04) by successfully deleting a protected admin table.


aliziel:
Thank you for the PR, our team is aware of the security considerations. Plans to update this workflow do not currently include community contributions, closing.

Issue #443, opened 2026-04-16
Airflow 3.0+ auth updates


ividito:
Airflow 3.0 reworks auth significantly. There are [simple upgrade paths ](https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html#breaking-changes)to carry our existing webserver functionality over to Airflow 3+ (maintaining keycloak functionality). A stretch goal here could be to investigate and implement new auth integrations, such as the new [Airflow/Keycloak provider](https://airflow.apache.org/docs/apache-airflow-providers-keycloak/stable/auth-manager/setup/config.html).

## AC

- [ ] Auth configuration updated to be Airflow 3-compatible
- [ ] Keycloak provider evaluated (either implemented or documented in a new issue for later)

Issue #442, opened 2026-04-16
Upgrade DAGs to be Airflow 3-compatible


ividito:
`uvx ruff check dags/ --select AIR301` shows 4 broken DAGs in 3.0.

```
AIR301 `airflow.utils.dates.days_ago` is removed in Airflow 3.0
   --> rds_disaster_recovery/get_snapshots_dag.py:258:45
help: Use `pendulum.today('UTC').add(days=-N, ...)` instead

AIR301 `airflow.utils.dates.days_ago` is removed in Airflow 3.0
   --> rds_disaster_recovery/rds_s3_export_snapshots.py:99:16
help: Use `pendulum.today('UTC').add(days=-N, ...)` instead

AIR301 `airflow.contrib.operators.slack_webhook_operator.SlackWebhookOperator` is removed in Airflow 3.0
  --> slack_notifications.py:26:18
help: The whole `airflow.contrib` module has been removed.

AIR301 [*] `schedule_interval` is removed in Airflow 3.0
  --> veda_data_pipeline/veda_stactools_noaa_hrrr.py:75:9
help: Use `schedule` instead
```

There may be other undetected issues - we should start by applying the recommended changes, then try to process our DAGs locally using `airflow>=3.0`

## AC
- [x] known deprecations resolved
- [x] DAGs tested locally with upgraded airflow version

## Stretch Goal

In Airflow 3, automatic DAGs where the first task short-circuits other tasks can be a lot noisier. If there's a change we can make to reduce this noise, that would be very helpful.

- [ ] Reduce "upstream failed" warnings when short-circuits are successful

ividito:
Will be closed with #455 

The stretch goal can be resolved with the new `@skip_if` decorator - we can defer this, since Airflow 3.1 cleans the default UI settings up a lot.

PR #449, opened 2026-05-14
Airflow 3.0 upgrade

### LINKED ISSUE #441: SM2A upgrade to Airflow 3.0+
> Direct DB access has been replaced by an intermediate service, and a new service is needed to run the DAG processor. We should be able to apply these changes in place, but the infra changes are significant enough to require some extra testing in a low-stakes environment.

Ref: https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html#understanding-airflow-3-x-architecture-changes

At first read, this looks like DB connections have been refactored and a new service has been added to facilitate worker -> DB connections. These changes will need to be applied at the module level for SM2A, and possibly in the Dockerfiles for each core service as well.

## AC

- [ ] Airflow 3.0+ is deployable using https://github.com/NASA-IMPACT/self-managed-apache-airflow/ - marked as a major version upgrade, since these are breaking changes.

ividito:
### Summary

Resolves #441 and #443. This branch also includes #442 (h/t @botanical)

### What changed

**Runtime & dependencies**

- Bump SM2A module to `v1.2.0-rc4`; Airflow 3.0 base images for `airflow_services` and `airflow_worker`.
- Move to the FAB auth manager (FAB is no longer bundled in Airflow 3) with OAuth/Keycloak.
- Make Keycloak public-key fetch lazy so the webserver no longer makes a network call at import time (and Keycloak import is skippable when not enabled).

**DAG code migration (Airflow 3 execution model)**
- `ti.dag_run` is no longer available to tasks — swept all DAGs to use the injected `dag_run` parameter instead.
- Migrate Dataset/DatasetAlias API to the Asset/AssetAlias API.

**Infrastructure**
- Use new SM2A module version to upgrade to Airflow 3 - most changes are actually in https://github.com/NASA-IMPACT/self-managed-apache-airflow/pull/38
- Set `AIRFLOW__API__BASE_URL` so workers can reach the Execution API
- Reduce API server to a single worker; tune dag-processor parse/refresh frequency, as the single worker was getting easily throttled.

## PR Checklist

- [x] Unit tests
- [x] Ad-hoc testing - *This is deployed and live in SIT*
- [x] Integration tests - https://gist.github.com/ividito/d0126cf74696c6ab4077e9e208af8c55
- [x] Infrastructure changes - test using `terraform validate` and `terraform plan`


infrastructure/main.tf
```diff
@@ -91,6 +91,38 @@ module "sma-base" {
         name  = "AIRFLOW__CORE__DEFAULT_TASK_RETRIES"
         value = var.workers_task_retries
       },
+      {
+        # Pin api-server to a single worker; workers > 1 crash-loops on Airflow 3.0.2.
+        name  = "AIRFLOW__API__WORKERS"
+        value = "1"
```
botanical, line 98: this is tea 🍵 Should we add a todo or something to fix this later?

sm2a-local-config/local_airflow.cfg
```diff
@@ -1,12 +1,17 @@
 [api]
-auth_backends = airflow.api.auth.backend.basic_auth
+auth_backends = airflow.providers.fab.auth_manager.api.auth.backend.basic_auth
+secret_key = Ly8wMU8r5K7jPy58M3GpkZbXDNyJz8HiJll3pu8DbIM=
+
+[api_auth]
+jwt_secret = Ly8wMU8r5K7jPy58M3GpkZbXDNyJz8HiJll3pu8DbIM=
```
botanical, line 6: do we need these secrets here?

airflow_services/requirements.txt
```diff
@@ -6,11 +7,13 @@ cryptography
 pycurl
 psycopg2-binary
 apache-airflow-providers-celery
-airflow_multi_dagrun
+apache-airflow-providers-fab
```
botanical: this is a duplicate of line 16 🔍 ?

dags/rds_disaster_recovery/get_snapshots_dag.py
```diff
@@ -312,11 +325,15 @@ def delete_glue_database_task(ti):
...
-        task_id="trigger_multi_s3_export_dag",
-        dag=dag,
-        trigger_dag_id="rds_s3_export_snapshots",
-        python_callable=trigger_s3_export_dag_task,
+    # rds_snapshots_dag_run = TriggerMultiDagRunOperator(
+    #     task_id="trigger_multi_s3_export_dag",
+    #     dag=dag,
+    #     trigger_dag_id="rds_s3_export_snapshots",
+    #     python_callable=trigger_s3_export_dag_task,
+    # )
```
botanical, line 333: We can delete the comments now right?

botanical (approved):
I had a couple small comments, non blocking ✅

infrastructure/main.tf
ividito: Probably not - we don't get enough traffic to Airflow to need more than one API worker

sm2a-local-config/local_airflow.cfg
ividito: They have to be _something_ valid to make local dev work. The actual secrets get generated by terraform and stored securely in AWS Secrets Manager.

dags/rds_disaster_recovery/get_snapshots_dag.py
ividito: I kept this as a reference for how the old operator was used - this DAG is the least-tested code, so I figured it might be useful. We can remove it though.

Issue #413, opened 2025-12-11
SM2A UI should display the currently deployed release


anayeaye:
## What
It is possible to deploy veda infrastructure manually, making it possible to deploy something other than what is recorded in the veda-deploy environment variables. We need to expose what version of the code is actually running to users.

## AC
- [ ] A user can see what release version is currently deployed from the UI

aliziel:
Noting for posterity: Airflow includes _theirs_ at the bottom, but labels aren't specific, so let's identify the VEDA source + SHAs well.

<img width="538" height="268" alt="Image" src="https://github.com/user-attachments/assets/e7a0c06c-8725-4a45-add5-ceb6ef18f8c3" />

But +1, this would be super helpful

PR #457, opened 2026-06-17
feat: add veda-data-airflow versioning to ui

sandrahoang686:
**Ticket:** https://github.com/NASA-IMPACT/veda-data-airflow/issues/413

**Note:**  This was a bit more involved and trickier than I thought it would be since I had to track where the HTML was coming from and had to opt for mounting a local version instead so we could add the extra version details for the footer. It looked like we weren't getting the git tags and versions already either so I had to update our git workflow and add these variables to terraform 🤔. The screen shot is just grabbing from my local env file. Let me know if this could have been done another way though?? 

**Changes**

* Added local version of `/templates/main.html` file to override the default version in `v2.10.5` (default version [here](https://github.com/apache/airflow/blob/2.10.5/airflow/www/templates/airflow/main.html)) and have docker mount this instead
   * Added `display_version` plugin so version variables can be referenced in template https://flask.palletsprojects.com/en/stable/templating/#context-processors
* Added `Get latest release tag` step in git workflow to grab the git tag and register with terraform so release versions are available to ECS task

Screenshot 2026-06-17 at 11 57 38 AM

### DIFF
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -109,6 +109,11 @@ jobs:
         with:
           lfs: "true"
           submodules: "recursive"
+          fetch-depth: 0
+
+      - name: Get latest release tag
+        id: get_tag
+        run: echo "tag=$(git describe --tags --abbrev=0 2>/dev/null || echo 'unknown')" >> $GITHUB_OUTPUT
 
       - name: Configure AWS Credentials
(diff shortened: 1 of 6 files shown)


sandrahoang686:
Closing in favor of https://github.com/NASA-IMPACT/veda-data-airflow/pull/468

PR #499, opened 2026-08-14
Fix: metrics fix + queue timeout increase

ividito:
Fix a deprecation in the metrics container. Also increase the default queue timeout - this was impacting backfill tasks, which were taking extra time as a result of migrating in the middle of a fires ingest. There will be an upstream PR to align the SQS settings with this new value - https://github.com/NASA-IMPACT/self-managed-apache-airflow/pull/40

The timeout change has already been applied to dev via the console.

### DIFF
--- a/infrastructure/configuration/celery_config.py
+++ b/infrastructure/configuration/celery_config.py
@@ -5,19 +5,10 @@ from airflow.providers.celery.executors.default_celery import DEFAULT_CELERY_CON
 # From here https://github.com/apache/airflow/issues/16163
 # DEFAULT_CELERY_CONFIG['task_acks_late'] = False
 
-# Must exceed the runtime of the longest task. If SQS makes a message visible
-# again while its task is still running, the message is redelivered and a second
-# worker tries to start the same task instance. On Airflow 3 the Task Execution
-# API rejects that second start with `invalid_state`, the Celery executor reports
-# the task as failed, and the scheduler then kills the healthy original run --
-# the task fails despite never having errored.
-#
-# NOTE: this alone is not sufficient. Because `predefined_queues` is set below,
-# kombu does not create the queue, and the SQS queue's own VisibilityTimeout
-# attribute governs redelivery. That attribute must be raised to match (it is
-# created at the AWS default of 30s by the sm2a module, which does not set
-# aws_sqs_queue.visibility_timeout_seconds).
-VISIBILITY_TIMEOUT = 21600  # 6h; SQS allows up to 43200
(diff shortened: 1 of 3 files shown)


infrastructure/configuration/celery_config.py
```diff
@@ -4,12 +4,17 @@
 
 # From here https://github.com/apache/airflow/issues/16163
 # DEFAULT_CELERY_CONFIG['task_acks_late'] = False
-# DEFAULT_CELERY_CONFIG['broker_transport_options']['visibility_timeout'] = 300
+
+# Must exceed the longest task runtime, or SQS redelivers mid-run and the duplicate
+# delivery fails the task on Airflow 3. The queue's own VisibilityTimeout attribute
+# must match, since predefined_queues means kombu does not create the queue.
+VISIBILITY_TIMEOUT = 1800  # 30m; SQS allows up to 43200
```
ividito, line 11: https://github.com/NASA-IMPACT/self-managed-apache-airflow/pull/40 is what addresses this through Terraform

scripts/put_airflow_worker_autoscaling_metric_data.py
```diff
@@ -178,7 +178,9 @@ def get_task_count_where_state(states: List[str]) -> int:
             session.query(func.sum(tasks_query.c.count))
             .join(DagModel, DagModel.dag_id == tasks_query.c.dag_id)
             .filter(
-                DagModel.is_active.is_(True),
+                # Airflow 3 removed the DagModel.is_active column; is_stale is its
+                # inverse (a DAG missing from its bundle is marked stale).
+                DagModel.is_stale.is_(False),
```
botanical, line 183: > Airflow 3 removed the DagModel.is_active column

Is this true? 
https://airflow.apache.org/docs/apache-airflow/3.0.2/_modules/airflow/models/dag.html#DAG.get_is_active

botanical (approved):
One question but non-blocking

scripts/put_airflow_worker_autoscaling_metric_data.py
ividito: 🤔 Not sure where that method gets used, but it looks like it returns `return not self.is_stale`. `DagModel.is_active` is definitely gone.

ividito:
Before merging, we'll need to make a release of SM2A off of https://github.com/NASA-IMPACT/self-managed-apache-airflow/pull/40, and upgrade to that version.

aliziel (approved):
Can reapprove after the release is cut in source + added 👍

anayeaye (approved):

PR #501, opened 2026-08-17
chore: run pre-commit

jjfrench:
**Summary:**

- Updates pre-commit / pyproject.toml ruff selects

## Changes

- built off #500 

Using primarily markdownlint-cli2 and ruff:
```toml
[tool.ruff]
line-length = 88

[tool.ruff.lint]
select = [
    "AIR",  # apache-airflow
    "E",    # pycodestyle errors
    "W",    # pycodestyle warnings
    "F",    # Pyflakes
    "I",    # isort
    "UP",   # pyupgrade
    "B",    # flake8-bugbear
    "C4",   # flake8-comprehensions
    "SIM",  # flake8-simplify
    "PTH",  # flake8-use-pathlib
    "RET",  # flake8-return
    "TID",  # flake8-tidy-imports
    "TC",   # flake8-type-checking
    "ISC",  # flake8-implicit-str-concat
    "ICN",  # flake8-import-conventions
]
ignore = [
  "BLE001",  # Blind Exceptions
]

[tool.ruff.format]
docstring-code-format = true
docstring-code-line-length = "dynamic"

[tool.markdownlint-cli2.config]
MD013 = false
MD033 = { allowed_elements = ["span"] }
```

The main configuration is `"AIR",  # apache-airflow`, which includes rules for Airflow 3 but also basic airflow rules.
>
Code | Name | Message | Fix/Status
-- | -- | -- | --
AIR001 | airflow-variable-name-task-id-mismatch | Task variable name should match the task_id: "{task_id}" | Rule has been stable since v0.0.271
AIR002 | airflow-dag-no-schedule-argument | DAG or @dag should have an explicit schedule argument | Rule has been stable since 0.13.0
AIR003 | airflow-variable-get-outside-task | Variable.get() outside of a task | 🧪Rule has been in preview since 0.15.6
AIR004 | airflow-task-branch-as-short-circuit | @task.branch can be replaced with @task.short_circuit | 🧪Rule has been in preview since 0.15.12
AIR201 | airflow-xcom-pull-in-template-string | Use the .output attribute on the task object for "{task_id}" instead of xcom_pull in a template string | 🧪Rule has been in preview since 0.15.11🛠️Automatic fix available
AIR202 | airflow-task-implicit-multiple-outputs | @task-decorated function relies on multiple_outputs inference | 🧪Rule has been in preview since 0.15.14🛠️Automatic fix available
[...]

PR #500, opened 2026-08-17
chore: update uv usage

jjfrench:
**Summary:**

Addresses https://github.com/NASA-IMPACT/veda-architecture/issues/833

## Changes
- updates github actions versions
- uses uv commands for remaining actions
- updates pre-commit (**does not run them**)
- updates dependabot

## PR Checklist

- [ ] Unit tests
- [ ] Ad-hoc testing - Deploy changes and test manually
- [ ] Integration tests
- [ ] Infrastructure changes - test using `terraform validate` and `terraform plan`

PR #506, opened 2026-08-25
feat: upgrade to 3.2.2 & add versioning to ui

sandrahoang686:
**Ticket:** https://github.com/NASA-IMPACT/veda-data-airflow/issues/413
**Changes:**
* Upgraded to apache airflow `v3.2.2` ([latest](https://github.com/apache/airflow/releases/tag/3.2.2)) because of this reported bug ([here](https://github.com/apache/airflow/issues/67145))and fix([here](https://github.com/apache/airflow/pull/67169)) 
* Added new step in git workflow to grab the git tag and register with terraform so release versions are available to ECS task

**Preview:**
Screenshot 2026-07-01 at 12 52 56 PM

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -143,6 +143,18 @@ module "sma-base" {
       {
         name  = "AWS_REQUEST_PAYER"
         value = var.aws_request_payer
+      },
+      {
+        name  = "VEDA_AIRFLOW_VERSION"
+        value = var.veda_airflow_version
+      },
+      {
+        name  = "GIT_SHA"
+        value = var.git_sha
+      },
+      {
+        name  = "AIRFLOW__WEBSERVER__INSTANCE_NAME"
+        value = "VEDA SM2A ${var.veda_airflow_version} (${var.git_sha})"
       }
     ],
     local.airflow_dag_variable_env_entries
--- a/.github/workflows/cicd.yml
+++ b/.github/workflows/cicd.yml
@@ -112,6 +112,11 @@ jobs:
         with:
           lfs: "true"
           submodules: "recursive"
+          fetch-depth: 0
+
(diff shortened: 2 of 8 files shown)


jjfrench (commented):
would also require updating the pyproject.toml and uv.lock file - I wonder if there's a better way to sync all the pins

infrastructure/main.tf
```diff
@@ -143,6 +143,18 @@ module "sma-base" {
...
         value = var.aws_request_payer
+      },
+      {
+        name  = "VEDA_AIRFLOW_VERSION"
+        value = var.veda_airflow_version
+      },
+      {
+        name  = "GIT_SHA"
+        value = var.git_sha
+      },
```
jjfrench, line 154: I'm not a huge fan of adding git sha's as tags to individual resources as it can cause unnecessary redeploys of resources

but I'm not familiar with the resources it's tagging or the intended behavior - so it might be okay

jjfrench (approved):
Changes look fine, noting that I'd like to get in #500 and #501 but I can always rebase

Not sure if you want another pair of eyes before merging

sandrahoang686:
> Changes look fine, noting that I'd like to get in https://github.com/NASA-IMPACT/veda-data-airflow/pull/500 and https://github.com/NASA-IMPACT/veda-data-airflow/pull/501 but I can always rebase

@jjfrench thanks for reviewing, you can definitely get those merged in first, we want this PR to sit for a bit because I believe we want `v3.0.2` to run for a bit before bumping. I just wanted to get this up to date :)

infrastructure/main.tf
sandrahoang686: This is a really good point actually, i'm gonna take a look 👀

PR #513, opened 2026-09-11
Add optional post-ingest table configuration to veda_ingest_vector

slesaad:
Adds an optional `table_config` block to `veda_ingest_vector` that applies index and statistics DDL to a vector collection after ingest.

## Why

`ogr2ogr` creates the GiST index on the geometry column and has no option for an index on any other column, so a collection with a datetime column has no index on it. Queries like `/items?datetime=...` with no bounding box fall back to a sequential scan. The only automatic index step today is `alter_datetime_add_indexes_eis()` (`handler.py:314`), hardcoded to fire only when the prefix starts with `EIS/` — this generalizes that.

Compounding it, `-overwrite` drops and recreates the table, so any index added by hand is destroyed on every re-ingest.

## Usage

```json
{
  "collection": "hms_smoke",
  "table_config": {
    "schema": "public",
    "indexes": [
      {"columns": ["datetime"], "method": "btree", "concurrently": true}
    ],
    "analyze": true
  }
}
```

Omit the key entirely to skip the step. Existing configs are unaffected.

## Why it runs after ingest

Building an index in one bulk sort is cheaper than maintaining it row by row while data loads. The new `configure_table` task sits downstream of `ingest_vector_task.expand(...)`, so Airflow runs it **once after every mapped instance completes** rather than once per file:

```
Start → discover → get_files → ingest_vector_task (mapped) → configure_table → invalidate_cloudfront → End
```

One consequence worth knowing, documented in the DAG doc and the pipeline README: if a backfill is split across **several DAG runs**, omit `table_config` from the intermediate runs and set it only on the last. Otherwise the index is created after the first run and every later chunk loads into an indexed table — the situation this ordering exists to avoid. Within a single run it is handled automatically.

## Design notes

- **`concurrently` defaults to true.** A blocking `CREATE INDEX` takes an `ACCESS EXCLUSIVE` lock; on a large table the Features API is serving that is a visible outage, not just a slow ingest. This requires autocommit, since `CREATE INDEX CONCURRENTLY` cannot run inside a transaction block.
- **Constrained schema rather than a SQL escape hatch.** These configs come from `s3://$EVENT_BUCKET/collections/*.json` and from manual triggers, so arbitrary SQL in them is hard to review. Statements are composed with `psycopg2.sql`: every schema, table, column, index and access-method name reaches Postgres as a driver-quoted identifier, and no config value is formatted into SQL text. Identifiers are also validated against a strict pattern so a malformed config fails before any statement runs. PEP 750 template strings would need Python 3.14 and psycopg 3; the project is pinned to Python 3.12 and psycopg2.
- **Deterministic index names, truncated to 63 characters.** Postgres truncates silently, so an over-long name would not match what `IF NOT EXISTS` looks for and the index would be rebuilt every run.
- **Invalid indexes are reported.** An interrupted `CONCURRENTLY` build leaves an INVALID index that `IF NOT EXISTS` treats as present, so it would never self-repair.
- **Requires an explicit `collection`.** With a per-file `id_template` there is no single table to configure, and the step is skipped with a warning.

## Testing

`build_statements()` is separated from execution so SQL generation is testable without a database, and `run_statements(conn, ...)` takes an open connection so the execution path can be pointed at any Postgres without AWS.

- **33 unit tests** — statement structure (including that no config value appears in the SQL text), identifier rejection (including `'a"; DROP TABLE users; --'`), method allowlisting, naming and truncation.
- **12 integration tests**, skipped unless `TEST_DATABASE_URL` is set so the default run stays offline. These cover what unit tests cannot: the exact rendered SQL, that the DDL is valid Postgres, that `CONCURRENTLY` works under the autocommit handling, that the task result is JSON-serializable for XCom, that a re-run is a genuine no-op, that `ANALYZE` populates `pg_stats`, and that the `pg_index` query finds an index left INVALID.

```bash
docker run -d --name veda-test-pg -p 55432:5432 \
    -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=postgres \
    postgis/postgis:16-3.4

TEST_DATABASE_URL=postgresql://postgres:postgres@localhost:55432/postgres \
    pytest tests/test_table_config_integration.py
```

All 12 pass against PostGIS 16. Also rehearsed against real data — 932 HMS smoke features with the datetime index dropped to simulate the post-ingest state, where `ogr2ogr` has created only the geometry GiST — and confirmed the planner uses the result rather than only that the DDL ran:

```
Index Scan using idx_hms_smoke_datetime on hms_smoke
```

Full suite: 100 passed, 12 skipped. The 3 failures on this branch are pre-existing `ModuleNotFoundError: No module named 'stactools'` and reproduce identically on a clean `dev` tree.

## Deliberately not included

- **`mode: replace|append`.** Table lifecycle is currently a side effect of whether someone put `-overwrite` or `-append` in `extra_flags`. Promoting it to an explicit field would make the destructive behaviour visible and let the DAG know whether indexes need rebuilding — but it changes behaviour for every existing collection and deserves its own review.
- **The `-append` fan-out race.** The pipeline README recommends `-append` to merge multiple files into one collection, but mapped ingest tasks run concurrently and race on table creation. Fixing it properly needs the table pre-created, which needs schema knowledge the DAG does not have generically.

Draft while the above two are decided.

### DIFF
--- a/dags/veda_data_pipeline/utils/vector_ingest/table_config.py
+++ b/dags/veda_data_pipeline/utils/vector_ingest/table_config.py
@@ -120,14 +120,44 @@ def find_invalid_indexes(cursor, collection: str, schema: str = "public") -> lis
     return [row[0] for row in cursor.fetchall()]
 
 
+def run_statements(conn, collection: str, table_config: dict) -> dict:
+    """Execute a table_config against an open connection.
+
+    Separated from secret lookup so this can be pointed at any Postgres -- a local
+    container, a test fixture -- without AWS.
+    """
+    statements = build_statements(collection, table_config)
+    if not statements:
+        print("No table_config provided, skipping table configuration")
+        return {"status": "skipped", "statements": []}
+
+    # CREATE INDEX CONCURRENTLY cannot run inside a transaction block.
+    conn.autocommit = True
+
+    with conn.cursor() as cursor:
+        for statement in statements:
+            print(f"Running: {statement}")
+            cursor.execute(statement)
+
+        invalid = find_invalid_indexes(
+            cursor, collection, table_config.get("schema", "public")
+        )
+        if invalid:
+            print(
+                f"WARNING: invalid indexes on {collection} (likely an interrupted "
+                f"CONCURRENTLY build); drop and recreate them: {invalid}"
+            )
+
+    return {"status": "success", "statements": statements}
+
+
 def apply_table_config(collection: str, table_config: dict, vector_secret_name: str) -> dict:
     """Run the table_config statements against the features database."""
     import psycopg2
 
     from veda_data_pipeline.utils.vector_ingest.handler import get_secret
 
-    statements = build_statements(collection, table_config)
-    if not statements:
+    if not build_statements(collection, table_config):
         print("No table_config provided, skipping table configuration")
         return {"status": "skipped", "statements": []}
 
@@ -138,24 +168,7 @@ def apply_table_config(collection: str, table_config: dict, vector_secret_name: 
         user=secrets["username"],
         password=secrets["password"],
     )
-    # CREATE INDEX CONCURRENTLY cannot run inside a transaction block.
-    conn.autocommit = True
-
     try:
-        with conn.cursor() as cursor:
-            for statement in statements:
-                print(f"Running: {statement}")
-                cursor.execute(statement)
-
-            invalid = find_invalid_indexes(
-                cursor, collection, table_config.get("schema", "public")
-            )
-            if invalid:
-                print(
-                    f"WARNING: invalid indexes on {collection} (likely an interrupted "
-                    f"CONCURRENTLY build); drop and recreate them: {invalid}"
-                )
+        return run_statements(conn, collection, table_config)
     finally:
         conn.close()
-
-    return {"status": "success", "statements": statements}
--- a/dags/veda_data_pipeline/utils/vector_ingest/table_config.py
+++ b/dags/veda_data_pipeline/utils/vector_ingest/table_config.py
@@ -1,0 +1,161 @@
+"""Apply post-ingest table configuration (indexes, statistics) to a vector collection.
+
+Indexes are deliberately created *after* the data is loaded: building a B-tree in one
+bulk sort is cheaper than maintaining it row by row during ingest. The task that calls
+this therefore sits downstream of the mapped ingest tasks, so it runs once after every
+chunk has landed rather than once per chunk.
+
+For a backfill split across several DAG runs, omit ``table_config`` from the conf of the
+intermediate runs and include it only on the last one -- otherwise the index exists while
+the remaining chunks are still loading, which is the case this ordering exists to avoid.
+
+``ogr2ogr`` already creates the GiST index on the geometry column, and has no option for
+an index on any other column. This covers the rest.
+"""
+
+import re
+
+# Postgres identifiers as produced by ogr2ogr: lower-case alphanumerics and underscores.
+# Anything else is rejected rather than escaped, so a surprising name fails loudly
+# instead of being quietly quoted into a statement.
+IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+# Index methods worth allowing here. `method` reaches SQL as a bare keyword, so it is
+# checked against this set rather than quoted.
+ALLOWED_INDEX_METHODS = frozenset({"btree", "gist", "brin", "gin", "hash", "spgist"})
+
+MAX_IDENTIFIER_LENGTH = 63  # Postgres truncates beyond this, silently
+
+
+class InvalidTableConfig(ValueError):
+    """Raised when a table_config block cannot be turned into safe SQL."""
+
+
+def _identifier(value: str, kind: str) -> str:
+    """Validate a Postgres identifier and return it double-quoted."""
+    if not isinstance(value, str) or not IDENTIFIER_RE.match(value):
+        raise InvalidTableConfig(
+            f"invalid {kind}: {value!r} -- expected letters, digits and underscores"
+        )
+    return f'"{value}"'
+
+
+def index_name(table: str, columns: list, method: str) -> str:
+    """Build a deterministic index name, so re-runs are genuine no-ops.
+
+    Truncated to Postgres' identifier limit; without this a long table plus several
+    columns silently produces a different name than `IF NOT EXISTS` looks for, and the
+    index gets rebuilt on every run.
+    """
+    name = f"idx_{table}_{'_'.join(columns)}"
+    if method != "btree":
+        name = f"{name}_{method}"
+    return name[:MAX_IDENTIFIER_LENGTH]
+
+
+def build_statements(collection: str, table_config: dict) -> list:
+    """Turn a table_config block into the SQL statements that apply it.
+
+    Kept separate from execution so the generated SQL can be tested without a database.
+    """
+    if not table_config:
+        return []
+
+    schema = table_config.get("schema", "public")
+    schema_sql = _identifier(schema, "schema")
+    table_sql = _identifier(collection, "collection")
+    qualified = f"{schema_sql}.{table_sql}"
+
+    statements = []
+    for index in table_config.get("indexes", []):
+        columns = index.get("columns")
+        if not columns:
+            raise InvalidTableConfig(f"index entry has no columns: {index!r}")
+
+        method = index.get("method", "btree").lower()
+        if method not in ALLOWED_INDEX_METHODS:
+            raise InvalidTableConfig(
+                f"unsupported index method {method!r}; "
+                f"expected one of {sorted(ALLOWED_INDEX_METHODS)}"
+            )
+
+        columns_sql = ", ".join(_identifier(column, "column") for column in columns)
+        name = index.get("name") or index_name(collection, columns, method)
+
+        # CONCURRENTLY avoids the ACCESS EXCLUSIVE lock that would otherwise block reads
+        # for the whole build -- on a large table being served by the features API that
+        # is a visible outage, not just a slow ingest.
+        concurrently = "CONCURRENTLY " if index.get("concurrently", True) else ""
+
+        statements.append(
+            f"CREATE INDEX {concurrently}IF NOT EXISTS {_identifier(name, 'index name')} "
+            f"ON {qualified} USING {method} ({columns_sql})"
+        )
+
+    if table_config.get("analyze", True):
+        # Without statistics the planner will not use the indexes just created.
+        statements.append(f"ANALYZE {qualified}")
+
+    return statements
+
+
+def find_invalid_indexes(cursor, collection: str, schema: str = "public") -> list:
+    """Return indexes left INVALID by an interrupted CONCURRENTLY build.
+
+    A failed `CREATE INDEX CONCURRENTLY` leaves an index behind that is never used by the
+    planner but is still maintained on write, and `IF NOT EXISTS` considers it present --
+    so the next run will not repair it. Surfacing the name is enough to act on.
+    """
+    cursor.execute(
+        """
+        SELECT i.relname
+        FROM pg_index x
+        JOIN pg_class i ON i.oid = x.indexrelid
+        JOIN pg_class t ON t.oid = x.indrelid
+        JOIN pg_namespace n ON n.oid = t.relnamespace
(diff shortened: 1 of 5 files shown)


dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,174 @@
...
+ALLOWED_INDEX_METHODS = frozenset({"btree", "gist", "brin", "gin", "hash", "spgist"})
+
+MAX_IDENTIFIER_LENGTH = 63  # Postgres truncates beyond this, silently
+
+
+class InvalidTableConfig(ValueError):
+    """Raised when a table_config block cannot be turned into safe SQL."""
+
+
+def _identifier(value: str, kind: str) -> str:
```
bitner, line 34: What library are you using to connect to the database? While the regex is likely ok, I would still avoid creating ANY sql using f-strings or general string formatting functions as much as  possible.

If you are using pscopg 3, the legacy way to do this would be to just create a sql.Identifier (https://www.psycopg.org/psycopg3/docs/api/sql.html#psycopg.sql.Identifier) and to build things up using sql.SQL, but with python PEP 750 Template Strings, you can now use the template syntax (https://www.psycopg.org/psycopg3/docs/basic/tstrings.html) and can address an identifier using {variable:i} in the template.

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,174 @@
...
+
+        columns_sql = ", ".join(_identifier(column, "column") for column in columns)
+        name = index.get("name") or index_name(collection, columns, method)
+
+        # CONCURRENTLY avoids the ACCESS EXCLUSIVE lock that would otherwise block reads
+        # for the whole build -- on a large table being served by the features API that
+        # is a visible outage, not just a slow ingest.
+        concurrently = "CONCURRENTLY " if index.get("concurrently", True) else ""
+
+        statements.append(
```
bitner, line 90: As above, really try to avoid using f-strings and SQL together. Traditionally, this has been kind of a pain to do at times, but now with the template string literals, it should be much easier to really avoid any use of f-strings and replace them with t-strings.

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
slesaad: psycopg2 2.9 on Python 3.12, so t-strings aren't available. Switched to `psycopg2.sql.SQL` + `sql.Identifier` composition instead (b08b53f), same pattern as `veda_vector_delete.py`. Regex is kept only to fail a malformed config early.

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
slesaad: Same here — no t-strings on psycopg2/3.12, so `CREATE INDEX` and `ANALYZE` are composed with `sql.SQL` + `sql.Identifier` in b08b53f, including the access method (`USING "btree"`). No f-strings left in the SQL path.

bitner: It might be worthwhile to change to the latest psycopg 3 anyway, there are a lot of usability and performance changes between psycopg 2 and 3 plus a lot of nicities like using the t"" strings.

dags/veda_data_pipeline/veda_vector_pipeline.py
```diff
@@ -74,6 +112,35 @@ def ingest_vector_task(payload):
...
+
+    Placed downstream of `ingest_vector_task.expand(...)` on purpose: Airflow runs a plain
+    task after *all* mapped instances complete, so indexes are built once on the finished
+    table rather than once per chunk. See utils/vector_ingest/table_config.py.
+    """
+    from veda_data_pipeline.utils.vector_ingest.table_config import apply_table_config
+
+    conf = dag_run.conf
+    table_config = conf.get("table_config")
+    if not table_config:
```
anayeaye, line 127: should we move this to the top of apply_config right after line 122 and return immediately without the status json? I am trying to understand why this task is not marked as skipped and I think it might be that the return `{"status": "skipped"}` is interpreted as a payload 🤔 or have you tried that and it also skips the downstream invalidate cloudfront task? Same for the table requirement on line 128.

For example:
https://github.com/NASA-IMPACT/veda-data-airflow/blob/eea93dca26360f9b4e09d66f85231d75142f60df/dags/veda_data_pipeline/veda_vector_pipeline.py#L80-L82

dags/veda_data_pipeline/veda_vector_pipeline.py
```diff
@@ -37,6 +37,35 @@
...
+        "indexes": [
+            {"columns": ["datetime"], "method": "btree", "concurrently": true}
+        ],
+        "analyze": true
+    }
+}
+```
+
+- Requires an explicit `collection`; with a per-file `id_template` there is no single
+  table to configure, and the step is skipped.
```
anayeaye: I don't understand this

dags/veda_data_pipeline/veda_vector_pipeline.py
```diff
@@ -37,6 +37,35 @@
...
+- Indexes are built after load on purpose -- one bulk sort rather than per-row
+  maintenance during ingest.
+- `concurrently` defaults to true so the build does not take an `ACCESS EXCLUSIVE` lock
+  on a table the Features API is serving.
+- `-overwrite` drops and recreates the table, so any index is destroyed on each ingest
```
anayeaye: Can we add context for this? I think it applies to the DAG extra_flags above (not as a key within the new table_config object). 
**Inserting data from multiple files** Do not use the `-overwrite` extra_flags argument because the operation drops and recreates the table, so any index is destroyed on each ingest and rebuilt by this step.

dags/veda_data_pipeline/veda_vector_pipeline.py
anayeaye: @ividito will this end the DAG and skip all following tasks or do you think invalidate cloudfront would still be evaluated?

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,192 @@
...
+# underscores, matching the lower-case names ogr2ogr produces. Names always reach Postgres
+# as quoted identifiers through psycopg2.sql; this check makes a malformed table_config
+# fail with a clear error before any statement runs, rather than partway through.
+IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+# Index access methods accepted in table_config. Postgres rejects an unknown method on its
+# own; checking here reports it before any statement runs.
+ALLOWED_INDEX_METHODS = frozenset({"btree", "gist", "brin", "gin", "hash", "spgist"})
+
+MAX_IDENTIFIER_LENGTH = 63  # Postgres truncates beyond this, silently
```
anayeaye, line 31: 👍

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,192 @@
...
+import re
+
+import psycopg2
+from psycopg2 import sql
+
+# Schema, table, column and index names accepted in table_config: letters, digits and
+# underscores, matching the lower-case names ogr2ogr produces. Names always reach Postgres
+# as quoted identifiers through psycopg2.sql; this check makes a malformed table_config
+# fail with a clear error before any statement runs, rather than partway through.
+IDENTIFIER_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
```
anayeaye, line 25: Nice!

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,192 @@
...
+        return []
+
+    schema = _validate_identifier(table_config.get("schema", "public"), "schema")
+    table = sql.Identifier(schema, _validate_identifier(collection, "collection"))
+
+    statements = []
+    for index in table_config.get("indexes", []):
+        columns = index.get("columns")
+        if not columns:
+            raise InvalidTableConfig(f"index entry has no columns: {index!r}")
```
anayeaye, line 78: Something to think about: if we fail the DAG during table config we will not continue to the invalidate cloudfront task. I am in favor of failing here I just wonder if we can re-think how we clear the cloudfront caching in this DAG

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+
+        # CONCURRENTLY avoids the ACCESS EXCLUSIVE lock that would otherwise block reads
+        # for the whole build -- on a large table being served by the features API that
+        # is a visible outage, not just a slow ingest.
+        concurrently = sql.SQL("CONCURRENTLY " if index.get("concurrently", True) else "")
+
+        statements.append(
+            sql.SQL(
+                "CREATE INDEX {concurrently}IF NOT EXISTS {name} "
```
anayeaye, line 99:
```suggestion
                "CREATE INDEX {concurrently} IF NOT EXISTS {name} "
```
does this work as written?

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
```diff
@@ -0,0 +1,192 @@
...
+            cursor.execute(statement)
+            rendered.append(text)
+
+        invalid = find_invalid_indexes(
+            cursor, collection, table_config.get("schema", "public")
+        )
+        if invalid:
+            print(
+                f"WARNING: invalid indexes on {collection} (likely an interrupted "
+                f"CONCURRENTLY build); drop and recreate them: {invalid}"
```
anayeaye, line 168: Does this mean the operator needs to take action? Do they delete the table?

anayeaye (commented):
Still reading tests but wanted to share comments early

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
slesaad: yes it works, because

`concurrently` is `sql.SQL("CONCURRENTLY ")`  with a trailing space, or `sql.SQL("")`

see [line 95](https://github.com/NASA-IMPACT/veda-data-airflow/pull/513/changes/BASE..cc510e26b5a28a7386277aaf1e6826228bb150e8#diff-3adcc19faa3b0a80e624d33f2c97141af9cba1e9fa6ccd1c3917eaf38266d245R95)

dags/veda_data_pipeline/utils/vector_ingest/table_config.py
slesaad: Yes, a human, and no, not the table — just the index.

An interrupted `CREATE INDEX CONCURRENTLY` leaves an `INVALID` index that `IF NOT EXISTS` treats as present, so re-running never repairs it.

The fix is `DROP INDEX CONCURRENTLY <name>`; then re-run the DAG.

dags/veda_data_pipeline/veda_vector_pipeline.py
slesaad: it looks like skipping the task will cascade to subsequent cloudfront invalidate unless we change the trigger rule for cloudfront (which we can set to run even if things fail)

dags/veda_data_pipeline/veda_vector_pipeline.py
slesaad: updated the docs

dags/veda_data_pipeline/veda_vector_pipeline.py
slesaad: updated the docs. basically what it means is that if we use `id_template` we don't have a single table but a bunch of tables based on the files present and in that case, there's no single table to "configure", so it's skipped.

eventually, maybe we want to apply the indices to all the tables - but not including that for now.

dags/veda_data_pipeline/veda_vector_pipeline.py
slesaad: but i did update it to move the import as well as as return bare instead.

slesaad: this PR will skip and keep cloudfront running but hasn't been tested in a deployed airflow so we can think about that later https://github.com/NASA-IMPACT/veda-data-airflow/pull/518

anayeaye (approved):

